PR #11, opened 2016-03-28
Add stats.py

sindhus:


boxoffice/extapi/stats.py
```diff
@@ -0,0 +1,35 @@
...
+from datetime import datetime as dt
+from os import environ
+import requests
+from urllib import urlencode
+import json
+from tabulate import tabulate
+
+def get_stats():
+    """ Query development database for stats on a particular ItemCollection """
+    init_for('dev')
```
shreyas-satish: This can't be `dev`.

boxoffice/extapi/stats.py
```diff
@@ -0,0 +1,35 @@
...
+        results.append([item.title, LineItem.query.join(Order).filter(LineItem.item == item, Order.status.in_(started)).count(), LineItem.query.join(Order).filter(LineItem.item == item, Order.status.in_(sold)).count(), LineItem.query.join(Order).filter(LineItem.item == item, Order.status.in_(cancelled)).count()])
+    # Prepend and append tilde for the stats to be published with raw text highlighting
+    stats = "```" + tabulate(results, headers=["Event","Initiated", "Sold", "Cancelled"]) + "```"
+    return stats
+
+def post_stats(stats):
+    """
+    Post stats to #customersupport via Slack WEBHOOK_URL
+    """
+    data={"channel": "#customersupport", "username": "boxoffice-stats", "icon_emoji": ":moneybag:", "text": stats}
```
shreyas-satish: These details need to be in the `Organization` instance. `Organization` has a `details` field, which accommodate all of this.

boxoffice/extapi/stats.py
```diff
@@ -0,0 +1,35 @@
...
+    response = requests.post(environ.get('WEBHOOK_URL'), data = json.dumps(data))
```
shreyas-satish: Needs to read from `app.config`.

shreyas-satish:
How is this supposed to be run? And how often? I'm not sure if it makes sense to run this as a cron job; the slack channel will be spammed with too many similar stats updates. My suggestion is to post it to the channel, every time there an order has been completed.

boxoffice/extapi/stats.py
sindhus: this is the data that gets passed to webhool URL?

boxoffice/extapi/stats.py
sindhus: If the webhook URL is exposed in app.config via being checked into git, anybody can post to our Slack channel.

shreyas-satish: config variables are read from `development.py` and `production.py` too, which aren't checked into Git.

sindhus:
@shreyas in the [newer PR](https://github.com/hasgeek/boxoffice/pull/27) `stats.py` is a file under `boxoffice/extapi` dir and called from line 201 in boxoffice as a RQ worker job every time a order is completed (just before you `send_receipt_mail`). If no channel is sent in the RQ job, then default is #test.

PR #15, opened 2016-03-31
Sort items in each category

vidya-ram:
Sort items in each category based on column 'seq'


vidya-ram:
Will do a new PR based on master

PR #17, opened 2016-03-31
Added support for price-based discounts

shreyas-satish:
An item's price can now be tagged with a discount_policy. This price will be considered independent of time constraints. Example use case: An event admin would like to provide a ticket at a special price regardless of _when_ they buy it, while maintaining the usual price tiers for other customers.

### DIFF
--- a/alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
+++ b/alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
@@ -1,0 +1,32 @@
+"""add discount_policy_id to price
+
+Revision ID: 45de268cd444
+Revises: 4d7f840202d2
+Create Date: 2016-03-31 15:29:51.897720
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '45de268cd444'
+down_revision = '4d7f840202d2'
+
+from alembic import op  # noqa
+import sqlalchemy as sa  # noqa
+import sqlalchemy_utils  # noqa
+
+
+def upgrade():
+    op.add_column('price', sa.Column('discount_policy_id', sqlalchemy_utils.types.uuid.UUIDType(), nullable=True))
+    op.create_unique_constraint('price_item_id_discount_policy_id_key', 'price', ['item_id', 'discount_policy_id'])
+    op.create_foreign_key('price_discount_policy_id_key', 'price', 'discount_policy', ['discount_policy_id'], ['id'])
+    op.alter_column('discount_policy', 'percentage',
+               existing_type=sa.INTEGER,
+               nullable=True)
+    op.add_column('discount_policy', sa.Column('is_price_based', sa.Boolean(), nullable=True))
+
+
+def downgrade():
+    op.drop_constraint('price_discount_policy_id_key', 'price', type_='foreignkey')
+    op.drop_constraint('price_item_id_discount_policy_id_key', 'price', type_='unique')
+    op.drop_column('price', 'discount_policy_id')
+    op.drop_column('discount_policy', 'is_price_based')
--- a/alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
+++ b/alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
@@ -10,13 +10,13 @@ Create Date: 2016-03-31 15:29:51.897720
 revision = '45de268cd444'
 down_revision = '4d7f840202d2'
 
-from alembic import op  # noqa
-import sqlalchemy as sa  # noqa
-import sqlalchemy_utils  # noqa
+from alembic import op
+import sqlalchemy as sa
+import sqlalchemy_utils
 
 
 def upgrade():
-    op.add_column('price', sa.Column('discount_policy_id', sqlalchemy_utils.types.uuid.UUIDType(), nullable=True))
+    op.add_column('price', sa.Column('discount_policy_id', sqlalchemy_utils.types.uuid.UUIDType(binary=False), nullable=True))
     op.create_unique_constraint('price_item_id_discount_policy_id_key', 'price', ['item_id', 'discount_policy_id'])
     op.create_foreign_key('price_discount_policy_id_key', 'price', 'discount_policy', ['discount_policy_id'], ['id'])
     op.alter_column('discount_policy', 'percentage',
--- a/alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
+++ b/alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
@@ -32,4 +32,4 @@ def downgrade():
     op.drop_column('discount_policy', 'is_price_based')
     op.alter_column('discount_policy', 'percentage',
                existing_type=sa.INTEGER,
-               nullable=True)
+               nullable=False)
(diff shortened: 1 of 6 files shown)


alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
```diff
@@ -0,0 +1,32 @@
...
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '45de268cd444'
+down_revision = '4d7f840202d2'
+
+from alembic import op  # noqa
+import sqlalchemy as sa  # noqa
+import sqlalchemy_utils  # noqa
```
jace: You don't need `# noqa` on imports you're using.

alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def upgrade():
+    op.add_column('price', sa.Column('discount_policy_id', sqlalchemy_utils.types.uuid.UUIDType(), nullable=True))
```
jace: Needs to be `sqlalchemy_utils.types.uuid.UUIDType(binary=False)`

alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
```diff
@@ -0,0 +1,32 @@
...
+    op.create_unique_constraint('price_item_id_discount_policy_id_key', 'price', ['item_id', 'discount_policy_id'])
+    op.create_foreign_key('price_discount_policy_id_key', 'price', 'discount_policy', ['discount_policy_id'], ['id'])
```
jace: Foreign keys should be suffixed `fkey`

alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
```diff
@@ -0,0 +1,32 @@
...
+    op.alter_column('discount_policy', 'percentage',
+               existing_type=sa.INTEGER,
+               nullable=True)
+    op.add_column('discount_policy', sa.Column('is_price_based', sa.Boolean(), nullable=True))
+
+
+def downgrade():
+    op.drop_constraint('price_discount_policy_id_key', 'price', type_='foreignkey')
```
jace: Suffix `fkey`

alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
```diff
@@ -0,0 +1,32 @@
...
+    op.drop_constraint('price_item_id_discount_policy_id_key', 'price', type_='unique')
+    op.drop_column('price', 'discount_policy_id')
+    op.drop_column('discount_policy', 'is_price_based')
```
jace: Alter column is missing in downgrade.

alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def upgrade():
+    op.add_column('price', sa.Column('discount_policy_id', sqlalchemy_utils.types.uuid.UUIDType(), nullable=True))
+    op.create_unique_constraint('price_item_id_discount_policy_id_key', 'price', ['item_id', 'discount_policy_id'])
+    op.create_foreign_key('price_discount_policy_id_key', 'price', 'discount_policy', ['discount_policy_id'], ['id'])
+    op.alter_column('discount_policy', 'percentage',
+               existing_type=sa.INTEGER,
+               nullable=True)
+    op.add_column('discount_policy', sa.Column('is_price_based', sa.Boolean(), nullable=True))
```
jace: This shouldn't be nullable.

boxoffice/models/line_item.py
```diff
@@ -102,12 +102,17 @@ def get_valid_discounts(self, line_items, coupons):
...
         item = Item.query.get(line_items[0].item_id)
-        if not coupons:
-            return DiscountPolicy.get_from_item(item, len(line_items))
         return DiscountPolicy.get_from_item(item, len(line_items), coupons)
 
-    def calculate_discounted_amount(self, percentage, base_amount):
-        return (percentage * base_amount/Decimal(100))
+    def calculate_discounted_amount(self, discount_policy, line_item):
+        if discount_policy.is_price_based:
+            item = Item.query.get(line_item.item_id)
```
jace: What's wrong with `item = line_item.item`?

boxoffice/models/line_item.py
```diff
@@ -122,10 +127,12 @@ def apply_discount(self, policy_coupon, line_items, combo=False):
         # unpack (discount_policy, dicount_coupon)
         discount_policy, coupon = policy_coupon
         for line_item in line_items:
-            discounted_amount = self.calculate_discounted_amount(discount_policy.percentage, line_item.base_amount)
-            if should_apply_discount and (not line_item.discount_policy_id or (combo and line_item.discounted_amount < discounted_amount)):
-                # if the discount policy's upper limit hasn't been reached,and if the line
-                # item hasn't been assigned a discount or if the line item's assigned discount is lesser
+            discounted_amount = self.calculate_discounted_amount(discount_policy, line_item)
+            if should_apply_discount and discounted_amount and \
```
jace: Move the `(` from the next line here so you don't need the `\`

boxoffice/models/line_item.py
```diff
@@ -122,10 +127,12 @@ def apply_discount(self, policy_coupon, line_items, combo=False):
...
+                (not line_item.discount_policy_id or (combo and line_item.discounted_amount < discounted_amount)):  # noqa
```
jace: Why does this need `# NOQA`?

boxoffice/models/line_item.py
```diff
@@ -122,10 +127,12 @@ def apply_discount(self, policy_coupon, line_items, combo=False):
...
+                # if the discount policy's upper limit hasn't been reached, and if discounted_amount
+                # is positive and if the line and if the item hasn't been assigned a discount
```
jace: "if the line and if the item"?

alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
shreyas-satish: The issue here is, setting `nullable` to `False` in the downgrade would cause the migration to fail, since some columns would have null values for `percentage`.

boxoffice/models/line_item.py
shreyas-satish: `line_item` here is a tuple, and only the `item_id` is assigned.

alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
shreyas-satish: Yes, will add another migration that makes this non-nullable.

alembic/versions/45de268cd444_add_discount_policy_id_to_price.py
jace: Downgrade is only for immediate reversal. You don't have to worry about downgrades after you start inserting data. However, database structure integrity is crucial, so you can't have migrations generating different database structures from fresh installations.

PR #22, opened 2016-04-06
Coupon limit

shreyas-satish:


alembic/versions/50f34bb47dc4_add_usage_limit_to_discount_coupon.py
```diff
@@ -0,0 +1,36 @@
...
+  column('id', sqlalchemy_utils.types.uuid.UUIDType()),
+  column('used', sa.Boolean()),
+  column('usage_limit', sa.Boolean()))
+
+line_item = table('line_item',
+  column('discount_coupon_id', sqlalchemy_utils.types.uuid.UUIDType()))
+
+
+def upgrade():
+    op.add_column('discount_coupon', sa.Column('usage_limit', sa.Integer(), nullable=True, server_default='1'))
```
jace: In the model you've specified a client-side default.

alembic/versions/50f34bb47dc4_add_usage_limit_to_discount_coupon.py
```diff
@@ -0,0 +1,36 @@
...
+from sqlalchemy.sql import table, column
+
+
+discount_coupon = table('discount_coupon',
+  column('id', sqlalchemy_utils.types.uuid.UUIDType()),
+  column('used', sa.Boolean()),
+  column('usage_limit', sa.Boolean()))
+
+line_item = table('line_item',
+  column('discount_coupon_id', sqlalchemy_utils.types.uuid.UUIDType()))
```
jace: This should include `binary=False`

alembic/versions/50f34bb47dc4_add_usage_limit_to_discount_coupon.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def upgrade():
+    op.add_column('discount_coupon', sa.Column('usage_limit', sa.Integer(), nullable=True, server_default='1'))
+    op.drop_column('discount_coupon', 'used')
+
+
+def downgrade():
+    op.add_column('discount_coupon', sa.Column('used', sa.Boolean(), nullable=False, server_default='False'))
```
jace: I don't think PostgreSQL understands `False`. Use `0`.

jace: You need to remove the server default on `used` after adding the column.

boxoffice/models/line_item.py
```diff
@@ -1,16 +1,18 @@
...
 
 __all__ = ['LineItem']
 
 
 class LINE_ITEM_STATUS(LabeledEnum):
-    CONFIRMED = (0, __("Confirmed"))
-    CANCELLED = (1, __("Cancelled"))
+    ADDED = (0, __("Added"))
+    CONFIRMED = (1, __("Confirmed"))
+    CANCELLED = (2, __("Cancelled"))
```
jace: What happens to all the existing rows in the database? These numbers can't change.

alembic/versions/50f34bb47dc4_add_usage_limit_to_discount_coupon.py
```diff
@@ -0,0 +1,36 @@
...
+from alembic import op
+import sqlalchemy as sa
+import sqlalchemy_utils
+from sqlalchemy.sql import table, column
+
+
+discount_coupon = table('discount_coupon',
+  column('id', sqlalchemy_utils.types.uuid.UUIDType()),
+  column('used', sa.Boolean()),
+  column('usage_limit', sa.Boolean()))
```
jace: This isn't boolean.

boxoffice/models/line_item.py
shreyas-satish: I was planning a data migration, but shall I commit make it an explicit migration?

boxoffice/models/line_item.py
```diff
@@ -73,6 +75,38 @@ def calculate(cls, line_item_dicts, coupons=[]):
             line_items.extend(item_line_items[item_id])
         return line_items
 
+    @classmethod
+    def coupon_used_count(cls, coupon):
+        count_query = select([func.count(LineItem.discount_coupon_id)]).where(LineItem.discount_coupon_id == coupon.id)
+        result = db.session.execute(count_query)
+        return result.first()[0]
```
jace: Why is this classmethod here? This should be an instance method on DiscountCoupon.

boxoffice/models/line_item.py
```diff
@@ -114,32 +148,32 @@ def calculate_discounted_amount(self, discount_policy, line_item):
             return line_item.base_amount - discounted_price
         return (discount_policy.percentage * line_item.base_amount/Decimal(100))
 
+    def is_coupon_usable(self, coupon, applied_to_count):
+        return (coupon.usage_limit - LineItem.coupon_used_count(coupon)) > applied_to_count
```
jace: This is inefficient. You are now running a query on every GET request. What you need instead:

1. A `used_count` column on the DiscountCoupon model
2. An `update_used_count` method in DiscountCoupon that runs `self.used_count = select([func.count(LineItem.discount_coupon_id)]).where(LineItem.discount_coupon_id == coupon.id)` (do not execute the query!)
3. A call to this method every time LineItem is updated or edited.

boxoffice/models/line_item.py
jace: No, don't change numbers. Made `Added` 2 and keep the existing ones.

alembic/versions/50f34bb47dc4_add_usage_limit_to_discount_coupon.py
```diff
@@ -0,0 +1,40 @@
...
+  column('used_count', sa.Integer()),
+  column('usage_limit', sa.Integer()))
+
+line_item = table('line_item',
+  column('discount_coupon_id', sqlalchemy_utils.types.uuid.UUIDType(binary=False)))
+
+
+def upgrade():
+    op.add_column('discount_coupon', sa.Column('usage_limit', sa.Integer(), nullable=True, server_default='1'))
+    op.add_column('discount_coupon', sa.Column('used_count', sa.Integer(), nullable=False, server_default='0'))
```
jace: `server_default` needs to be removed.

boxoffice/models/line_item.py
```diff
@@ -73,6 +75,39 @@ def calculate(cls, line_item_dicts, coupons=[]):
...
+            if coupon and coupon.usage_limit > coupon.used_count:
+                policies.append((coupon.discount_policy, coupon))
+
+    return policies
+
+DiscountPolicy.get_from_item = classmethod(get_from_item)
+
+
+def update_used_count(self):
+    self.used_count = select([func.count(LineItem.discount_coupon_id)]).where(LineItem.discount_coupon_id == self.id).where(LineItem.status == LINE_ITEM_STATUS.CONFIRMED).as_scalar()
```
jace: I'd prefer to not use `id` columns for comparison so that if there's a type mismatch in your code, SQLAlchemy will catch it for you. In this case, this should be:

`self.used_count = select([func.count()]).where(LineItem.discount_coupon == self).where(LineItem.status == LINE_ITEM_STATUS.CONFIRMED).as_scalar()`

PR #23, opened 2016-04-06
UI change when discount coupon is applied

vidya-ram:
Add corresponding ticket to cart when discount coupon is applied.


vidya-ram:
Closing it. Will send a new branch based out of master.

PR #28, opened 2016-04-11
Update .gitignore

sindhus:
Now ignores Baseframe cache and assets


jace:
Can you also `git rm` the `baseframe-packed.*` files?

PR #29, opened 2016-04-12
Participant assignment

shreyas-satish:


alembic/versions/253e7b76eb8e_modify_assignee.py
```diff
@@ -0,0 +1,40 @@
...
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '253e7b76eb8e'
+down_revision = '1ea1e8070ac8'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
```
jace: Unused import

alembic/versions/253e7b76eb8e_modify_assignee.py
```diff
@@ -0,0 +1,40 @@
...
+import sqlalchemy_utils
+
+
+def upgrade():
+    op.add_column('assignee', sa.Column('current', sa.Boolean(), nullable=True))
```
jace: Add a check constraint to ensure the value is never `false`. That leaves `null` and `true` as the permissible values.

boxoffice/models/line_item.py
```diff
@@ -26,21 +26,23 @@ def make_ntuple(item_id, base_amount, **kwargs):
...
     #: Unvalidated email address
     email = db.Column(db.Unicode(254), nullable=False)
     #: Unvalidated phone number
     phone = db.Column(db.Unicode(16), nullable=True)
     details = db.Column(JsonDict, nullable=False, default={})
-
-    # Track the assignee from whom the line_item was transferred from
-    previous_id = db.Column(None, db.ForeignKey('assignee.id'), nullable=True)
-    previous = db.relationship('Assignee', uselist=False)
+    current = db.Column(db.Boolean, default=False, nullable=True)
```
jace: No default here. `False` is an invalid value. (The default for `default` is `None`, which is what you want.)

boxoffice/models/line_item.py
```diff
@@ -100,6 +99,10 @@ def calculate(cls, line_item_dicts, coupons=[]):
     def confirm(self):
         self.status = LINE_ITEM_STATUS.CONFIRMED
 
+    @property
+    def current_assignee(self):
+        return self.assignees.filter(Assignee.current == True).one_or_none()
```
jace: This can be a relationship: `assignee = db.relationship(Assignee, primaryjoin=Assignee.line_item == self and Assignee.current == True)`

Don't use `current_assignee` -- we want to imply that there can only be one assignee and the rest are historical (and hence not 'assignees')

boxoffice/views/order.py
```diff
@@ -213,3 +245,12 @@ def payment(order):
...
     return render_template('cash_receipt.html', order=order, org=order.organization, line_items=line_items)
+
+
+@app.route('/order/<access_token>/ticket', methods=['GET', 'POST'])
+@render_with({'text/html': 'order.html', 'application/json': jsonify_order}, json=True)
+@load_models(
+    (Order, {'access_token': 'access_token'}, 'order')
+    )
+def line_items(order):
+    return dict(order=order, org=order.organization)
```
jace: You need to verify the access token here.

boxoffice/views/participant.py
```diff
@@ -0,0 +1,36 @@
+from flask import request, jsonify, make_response
+from .. import app
+from ..models import db
+from ..models import LineItem, Assignee
+from ..models import Order
+from coaster.views import load_models
+from utils import xhr_only
+
+
+@app.route('/participant/<order>/assign', methods=['GET', 'OPTIONS', 'POST'])
```
jace: Access token missing.

boxoffice/views/participant.py
```diff
@@ -0,0 +1,36 @@
...
+@load_models(
+    (Order, {'id': 'order'}, 'order')
+    )
+@xhr_only
+def assign(order):
+    """
+    Assign a line_item to a participant
+    """
+    if not request.json or not request.json.get('attendee'):
```
jace: Access token verification missing.

alembic/versions/253e7b76eb8e_modify_assignee.py
```diff
@@ -0,0 +1,41 @@
...
+down_revision = '1ea1e8070ac8'
+
+from alembic import op
+import sqlalchemy as sa
+import sqlalchemy_utils
+
+
+def upgrade():
+    op.add_column('assignee', sa.Column('current', sa.Boolean(), nullable=True))
+    op.create_check_constraint('assignee_true_or_none_check', 'assignee', u"current != 'f'")
```
jace: Shouldn't this be `0` not `'f'`?

jace: And since the check is on the `current` column, the constraint should be `assignee_current_check`.

boxoffice/models/line_item.py
```diff
@@ -26,21 +26,24 @@ def make_ntuple(item_id, base_amount, **kwargs):
 
 class Assignee(BaseMixin, db.Model):
     __tablename__ = 'assignee'
+    __table_args__ = (db.UniqueConstraint('line_item_id', 'current'),
+        db.CheckConstraint("current != '0'", 'assignee_current_check'))
```
jace: no quotes around the `'0'` here.

alembic/versions/253e7b76eb8e_modify_assignee.py
```diff
@@ -0,0 +1,41 @@
...
+down_revision = '1ea1e8070ac8'
+
+from alembic import op
+import sqlalchemy as sa
+import sqlalchemy_utils
+
+
+def upgrade():
+    op.add_column('assignee', sa.Column('current', sa.Boolean(), nullable=True))
+    op.create_check_constraint('assignee_current_check', 'assignee', u"current != '0'")
```
jace: Still quoted?

PR #35, opened 2016-04-13
added basic api for orders

shreyas-satish:


boxoffice/models/line_item.py
```diff
@@ -107,6 +107,12 @@ def current_assignee(self):
         return self.assignees.filter(Assignee.current == True).one_or_none()
 
 
+@property
+def get_confirmed_line_items(self):
+    return LineItem.query.filter(LineItem.order == self, LineItem.status == LINE_ITEM_STATUS.CONFIRMED).all()
+Order.get_confirmed_line_items = get_confirmed_line_items
```
jace: Don't use the @property decorator. Instead, use `Order.get_confirmed_line_items = property(get_confirmed_line_items)`. Also, two space above.

boxoffice/views/order.py
```diff
@@ -261,3 +261,43 @@ def receipt(order):
...
+    return api_orders
+
+
+@app.route('/api/1/ic/<item_collection>/orders', methods=['GET', 'OPTIONS'])
+@load_models(
+    (ItemCollection, {'id': 'item_collection'}, 'item_collection')
+    )
+def item_collection_orders(item_collection):
+    organization = item_collection.organization
+    if not request.args.get('access_token') or request.args.get('access_token') != organization.details.get("access_token"):
```
jace: This is a hack. Okay for now, but needs to be replaced ASAP.

shreyas-satish: Sure, added a comment and filing an issue.

PR #40, opened 2016-05-24
Basic dashboard

shreyas-satish:
This pull request introduces a basic dashboard, which gives an overview of the registrations and sales for an item collection.

boxoffice_dash

The dashboard is architected as a single-page app. The app uses a Model-View-Template structure (based on the Model-View-Controller pattern). The Model is supposed to handle the business of fetching, caching and validating data. The View, invoked by the Router, manages the control flow, while the Template is the HTML template (with dynamic Ractive variables).

The following technologies have been used:
1. Backbone.js for models, the router and the event dispatcher.
2. Ractive for reactive templating.
3. ES6 as the default language. This is transpiled with babel to ES5.
4. Gulp as the build tool.
5. Browserify for bundling up modules.
6. d3 and c3 for charts.
7. underscore, which is a backbone dependency, as a utility library.
8. jQuery, which is internally used by Backbone to handle ajax calls.

For development setup:

1. `cd boxoffice/static/js`
2. `npm install`
3. `npm install -g gulp`
4. `gulp watch` to watch and bundle the JS files as they change.


boxoffice/models/item_collection.py
```diff
@@ -1,3 +1,4 @@
+from sqlalchemy import func
```
jace: Unused import

boxoffice/models/line_item.py
```diff
@@ -106,6 +106,39 @@ def confirm(self):
...
+        """
+        Returns number of line items sold per date per item.
+        Eg: {'2016-01-01': {'item-xxx': 20}}
+        """
+        date_item_counts = {}
+        for item in item_collection.items:
+            item_id = unicode(item.id)
+            item_results = db.session.query(func.date(cls.ordered_at), func.count(cls.id)).filter(
+                cls.item == item, cls.status == LINE_ITEM_STATUS.CONFIRMED).order_by(
+                func.date(cls.ordered_at).asc()).group_by(func.date(cls.ordered_at)).all()
```
jace: This is not accounting for the user's timezone. Dates are stored in UTC.

boxoffice/templates/order.html
```diff
@@ -139,4 +139,5 @@ <h4 class="text-center attendee-form-title">Please fill the attendee details</h4
   </div>
 {% endblock %}
 {% block footerscripts %}
+  <script src='/static/js/views/order.js'></script>
```
vidya-ram: In this case, then in file boxoffice/views/__init__.py
js assets will have to be updated.
assets['boxoffice.js'][version] = 'js/scripts.js'

boxoffice/static/js/templates/item_collection.html.js
```diff
@@ -0,0 +1,57 @@
...
+        {{#items}}
+          <tr>
+            <td>{{ @index + 1 }}</td>
+            <td>{{ title }}</td>
+            <td>{{ available }}</td>
+            <td>{{ sold }}</td>
+            <td>{{ free }}</td>
+            <td>{{ cancelled }}</td>
+            <td>{{ current_price }}</td>
+            <td>{{ net_sales }}</td>
```
vidya-ram: Better to include currency symbol

boxoffice/static/js/templates/item_collection.html.js
```diff
@@ -0,0 +1,57 @@
...
+export const ItemCollectionTemplate = `
+  <br>
+  <div class="row">
+    <div class="col-md-4">
+      <div class="panel panel-default">
+        <div class="panel-heading">
+          <h3 class="panel-title">Net Sales</h3>
+        </div>
+        <div class="panel-body">
+          {{net_sales}}
```
vidya-ram: Better to include currency symbol

shreyas-satish:
Updated UI.

boxoffice_31516

boxoffice/models/line_item.py
```diff
@@ -106,6 +107,53 @@ def confirm(self):
     def current_assignee(self):
         return self.assignees.filter(Assignee.current == True).one_or_none()
 
+    @classmethod
+    def counts_per_date_per_item(cls, item_collection, timezone):
```
jace: Why is this a classmethod here? It does not even refer to the class.

boxoffice/models/line_item.py
```diff
@@ -106,6 +107,53 @@ def confirm(self):
...
+            ).params(timezone=timezone, status=LINE_ITEM_STATUS.CONFIRMED, item_id=item.id)
+            for res in item_results:
+                if not date_item_counts.get(res[0].isoformat()):
+                    date_item_counts[res[0].isoformat()] = {item_id: res[1]}
+                else:
+                    date_item_counts[res[0].isoformat()][item_id] = res[1]
+        return date_item_counts
+
+    @classmethod
+    def sales_by_date(cls, dates, user_tz):
```
jace: This shouldn't be a class method either.

boxoffice/models/line_item.py
```diff
@@ -106,6 +107,53 @@ def confirm(self):
...
+        date_sales = {}
+        for sales_date in dates:
+            date_sale_res = db.session.query('sum').from_statement('''SELECT SUM(final_amount) FROM line_item
+                WHERE status=:status AND DATE_TRUNC('day', line_item.ordered_at AT TIME ZONE 'UTC' AT TIME ZONE :timezone)::date = :sales_date
+                ''').params(timezone=user_tz, status=LINE_ITEM_STATUS.CONFIRMED, sales_date=sales_date).first()
+            date_sales[sales_date] = date_sale_res[0] or Decimal(0)
+        return date_sales
+
+    @classmethod
+    def sales_delta(cls, user_tz):
```
jace: This too shouldn't be a class method.

boxoffice/models/user.py
```diff
@@ -14,6 +14,10 @@ class User(UserBase2, db.Model):
     def __repr__(self):
         return self.fullname
 
+    @property
+    def orgs(self):
```
jace: Make this `organizations_owned`, and a function, not a property. It has a non-trivial call cost.

jace: Alternatively, you can leave it as a property, but without the `.all()` call.

boxoffice/models/user.py
```diff
@@ -25,3 +29,9 @@ class Organization(ProfileBase, db.Model):
 
     details = db.Column(JsonDict, nullable=False, server_default='{}')
     contact_email = db.Column(db.Unicode(254), nullable=False)
+
+    def permissions(self, user, inherited=None):
+        perms = super(Organization, self).permissions(user, inherited)
+        if self.userid in user.organizations_owned_ids():
+            perms.update(['org_admin'])
```
jace: `perms.add` more straightforward, no?

PR #45, opened 2016-06-09
added server-side validation and client-side errors to prevent over booking

shreyas-satish:


boxoffice/views/order.py
```diff
@@ -122,6 +125,11 @@ def order(item_collection):
     if not buyer_form.validate():
         return make_response(jsonify(message='Invalid buyer details'), 400)
 
+    for line_item_form in line_item_forms:
+        item = Item.query.get(line_item_form.data.get('item_id'))
+        if (item.get_confirmed_line_items.count() + line_item_form.data.get('quantity')) > item.quantity_total:
+            return make_response(jsonify(error_type='order_calculation', message='The selected quantity for {item} is not available. Please edit the order and update the quantity.'.format(item=item.title)), 400)
```
jace: This block is very db-inefficient. It makes two separate database calls for every line item. This could be entirely one call.

jace: In the error message, place `{item}` in quotes. Also, why are these strings not wrapped in gettext?

boxoffice/models/line_item.py
```diff
@@ -108,6 +108,23 @@ def current_assignee(self):
...
 
+def get_availability(cls, item_ids):
+    """Returns a dict -> {'item_id': ('item title', 'quantity_total', 'line_item_count')}"""
+    items_dict = {}
+    item_tups = db.session.query(cls.id, cls.title, cls.quantity_total, func.count(cls.id)).join(LineItem).filter(
+        LineItem.item_id.in_(item_ids), LineItem.status == LINE_ITEM_STATUS.CONFIRMED).group_by(cls.id).all()
+    for item_tup in item_tups:
+        items_dict[unicode(item_tup[0])] = item_tup[1:]
+    return items_dict
+Item.get_availability = classmethod(get_availability)
```
jace: Needs two blank lines above.

boxoffice/models/line_item.py
```diff
@@ -108,6 +108,23 @@ def current_assignee(self):
...
+
+
+def get_confirmed_line_items(self):
+    """Returns a SQLAlchemy query object preset with an item's confirmed line items"""
+    return LineItem.query.filter(LineItem.item == self, LineItem.status == LINE_ITEM_STATUS.CONFIRMED)
+Item.get_confirmed_line_items = property(get_confirmed_line_items)
```
jace: Two blank lines above.

boxoffice/models/line_item.py
```diff
@@ -108,6 +108,23 @@ def current_assignee(self):
         return self.assignees.filter(Assignee.current == True).one_or_none()
 
 
+def get_availability(cls, item_ids):
+    """Returns a dict -> {'item_id': ('item title', 'quantity_total', 'line_item_count')}"""
+    items_dict = {}
+    item_tups = db.session.query(cls.id, cls.title, cls.quantity_total, func.count(cls.id)).join(LineItem).filter(
+        LineItem.item_id.in_(item_ids), LineItem.status == LINE_ITEM_STATUS.CONFIRMED).group_by(cls.id).all()
+    for item_tup in item_tups:
+        items_dict[unicode(item_tup[0])] = item_tup[1:]
```
jace: Use named attributes for clarity. `item_tup.id` here.

boxoffice/models/line_item.py
shreyas-satish: Fixed.

boxoffice/models/line_item.py
shreyas-satish: Fixed.

boxoffice/models/line_item.py
```diff
@@ -108,6 +108,23 @@ def current_assignee(self):
         return self.assignees.filter(Assignee.current == True).one_or_none()
 
 
+def get_availability(cls, item_ids):
+    """Returns a dict -> {'item_id': ('item title', 'quantity_total', 'line_item_count')}"""
+    items_dict = {}
+    item_tups = db.session.query(cls.id, cls.title, cls.quantity_total, func.count(cls.id)).join(LineItem).filter(
```
jace: Make that `func.count(cls.id).alias('count')`

boxoffice/templates/boxoffice.html
```diff
@@ -39,11 +39,15 @@ <h1 class="category-heading">{{title}}</h1>
...
-                        <div class="ticket-description">
-                          <p class="few-ticket">Only a few tickets left</p>
-                        </div>
-                      {{/if}}
+                      <div class="ticket-description">
+                        <p class="few-ticket">
+                          {{#if line_item.quantity >= line_item.quantity_available}}
+                            Only {{line_item.quantity_available}} available
+                          {{elseif line_item.quantity_available > 0 && line_item.quantity_available < 5}}
+                            Only a few tickets left
```
jace: We need i18n in JS templates as well.

boxoffice/views/item_collection.py
```diff
@@ -13,7 +13,7 @@ def jsonify_item(item):
             'title': item.title,
             'id': item.id,
             'description': item.description.text,
-            'quantity_available': item.quantity_available,
+            'quantity_available': item.quantity_total - item.get_confirmed_line_items.count(),
```
jace: Why is `item.quantity_available` unreliable here, and why is that column present if it is not reliable?

boxoffice/models/line_item.py
shreyas-satish: I'm afraid I can't do that. `item_tup` is a tuple with values, eg: `(UUID('17e37e78-178f-11e6-b989-21e4c39b7dd7'), u'T-shirt', 6, 5L)`.

boxoffice/views/order.py
```diff
@@ -26,13 +28,16 @@ def jsonify_line_items(line_items):
     """
     items_json = dict()
     for line_item in line_items:
+        item = Item.query.get(line_item.item_id)
+        quantity_available = item.quantity_total - item.get_confirmed_line_items.count()
```
jace: As above, why is `item.quantity_available` not being used?

shreyas-satish: `item.quantity_available` was originally introduced with the intention of letting an admin manually update the value. But now, it seem simpler to only have `item.quantity_total` and compute the availability.

PR #52, opened 2016-06-22
Change buyer form UI

vidya-ram:
1. Change buyer form to google's material design UI and remove bootstrap form dependency
2. Change the UI to show the validation error message next to the input field.
3. Remove unused css.
4. Change the attendee assignment UI to show the validation error message next to the input field.
screen shot 2016-07-18 at 4 57 10 pm


boxoffice/templates/boxoffice.html
```diff
@@ -111,24 +111,27 @@ <h1 class="category-heading">{{title}}</h1>
...
           </div>
-          <div class="form-group phone-number-wrapper">
-            <div class="col-md-6 col-md-offset-3 col-sm-8 col-sm-offset-2">
-              <input type="text" class="form-control buyer-fields secondary" name="phone" value = {{buyer.phone }} placeholder="Phone"/>
-            </div>
+          <div class="group">
+            <input class="group-input {{#buyer.phone}}filled{{/}}" type="text" name="phone" value="{{buyer.phone}}">
+            <span class="bar"></span>
+            <label class="group-label">Phone</label>
+            {{#errormsg.phone}}<p class="form-error-msg">{{ errormsg.phone }}</p>{{/}}
```
shreyas-satish: Would a hint help? Such as `Eg: +917676332020`.

boxoffice/templates/boxoffice.js
```diff
@@ -242,7 +242,7 @@ $(function() {
               var line_items = boxoffice.ractive.get('order.line_items');
               line_items.forEach(function(line_item) {
                 if (data.line_items.hasOwnProperty(line_item.item_id)) {
-                  if(data.line_items[line_item.item_id].discounted_amount && line_item.quantity_available > 0) {
+                  if (data.line_items[line_item.item_id].discounted_amount && line_item.quantity_available > 0) {
                     valid_discount_coupon = true;
```
shreyas-satish: Why is `valid_discount_coupon` required?

boxoffice/templates/boxoffice.js
```diff
@@ -254,7 +254,7 @@ $(function() {
                 }
               });
 
-              if(valid_discount_coupon) {
+              if (valid_discount_coupon) {
                 boxoffice.ractive.set('order.line_items',line_items);
```
shreyas-satish: Is this `set` required? Setting `pre_applied` on the `discount_policy` should re-render the UI?

boxoffice/templates/boxoffice.js
vidya-ram: Depending on the value of `valid_discount_coupon`, `order.line_items` is set and page is scrolled to the Boxoffice widget (Line: 257)

boxoffice/templates/boxoffice.js
vidya-ram: Incase discount policy of the coupon is applicable to multiple items, wouldn't it be better to set `order.line_items` than setting each line_item.

shreyas-satish: Not necessarily, no? Since the line items that have a pre-applied discount will be changed anyway. I don't think `valid_discount_coupon` is required at all. Also, `valid_discount_coupon` as a variable name is slightly mis-leading, since it actually is just set if a line item has a discount applied to it.

PR #55, opened 2016-06-28
added transaction_ref, and bank_transfer enum

shreyas-satish:


boxoffice/models/payment.py
```diff
@@ -12,6 +12,7 @@
 class TRANSACTION_METHOD(LabeledEnum):
     ONLINE = (0, __("Online"))
     CASH = (1, __("Cash"))
+    BANK_TRANSFER = (2, __("Bank Transfer"))
```
jace: Where are you recording which payment gateway was used for this?

PR #67, opened 2016-07-06
Adds a sequence to Item

shreyas-satish:
The `seq` in `Item` is added to allow sequencing the items in a particular category or item collection. For instance, this `seq` can be used to list the items in an order as per marketing requirements.

### DIFF
--- a/alembic/versions/74770336785_add_seq_to_item.py
+++ b/alembic/versions/74770336785_add_seq_to_item.py
@@ -30,7 +30,7 @@ def upgrade():
     item_collections = connection.execute(sa.select([item_collection.c.id]))
     item_collection_ids = [ic_tuple[0] for ic_tuple in item_collections]
     for item_collection_id in item_collection_ids:
-        items = connection.execute(sa.select([item.c.id]).where(item.c.item_collection_id == item_collection_id))
+        items = connection.execute(sa.select([item.c.id]).where(item.c.item_collection_id == item_collection_id).order_by('created_at'))
         item_ids = [item_tuple[0] for item_tuple in items]
         for idx, item_id in enumerate(item_ids):
             op.execute(item.update().where(item.c.id == item_id).values({'seq': idx + 1}))
--- a/alembic/versions/74770336785_add_seq_to_item.py
+++ b/alembic/versions/74770336785_add_seq_to_item.py
@@ -12,10 +12,29 @@ down_revision = '59d274a1682f'
 
 from alembic import op
 import sqlalchemy as sa
+from sqlalchemy.sql import table, column
+import sqlalchemy_utils
+
+item_collection = table('item_collection', column('id', sqlalchemy_utils.types.uuid.UUIDType()))
+
+item = table('item',
+  column('id', sqlalchemy_utils.types.uuid.UUIDType()),
+  column('item_collection_id', sqlalchemy_utils.types.uuid.UUIDType()),
+  column('seq', sa.Integer()),
+  column('created_at', sa.DateTime()))
 
 
 def upgrade():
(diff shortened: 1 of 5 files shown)


boxoffice/views/item_collection.py
```diff
@@ -29,7 +29,7 @@ def jsonify_item(item):
 
 def jsonify_category(category):
     category_items = []
-    for item in category.items:
+    for item in Item.query.filter(Item.category == category).order_by('seq ASC').all():
```
jace: `ASC` is implicit. Don't need to specify it. Also, since this is a view, add a class method to `Item` that accepts the `category` parameter and returns the query. Don't use `.all()`. Queries are iterable by default, `all()` just casts the iterator into a list.

boxoffice/views/item_collection.py
```diff
@@ -60,8 +60,7 @@ def boxofficejs():
 @cors
 def item_collection(item_collection):
     categories_json = []
-    item_collection.categories.sort(key=lambda category: category.seq)
-    for category in item_collection.categories:
+    for category in Category.query.filter(Category.item_collection == item_collection).order_by('seq ASC').all():
```
jace: As above.

boxoffice/models/item.py
```diff
@@ -31,6 +31,7 @@ class Item(BaseScopedNameMixin, db.Model):
     assignee_details = db.Column(JsonDict, default={}, nullable=False)
 
     cancellable_until = db.Column(db.DateTime, nullable=True)
+    seq = db.Column(db.Integer, nullable=True)
```
jace: `seq` should be non-nullable. SQLAlchemy provides a collection class that you can use to automatically add numbers via `ItemCollection`. See Hascore's NetworkbarLink for how to do this.

alembic/versions/74770336785_add_seq_to_item.py
```diff
@@ -0,0 +1,22 @@
...
+# revision identifiers, used by Alembic.
+revision = '74770336785'
+down_revision = '59d274a1682f'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('item', sa.Column('seq', sa.Integer(), nullable=True))
```
jace: Add seq numbers and make this non-nullable.

alembic/versions/74770336785_add_seq_to_item.py
```diff
@@ -0,0 +1,41 @@
...
+  column('id', sqlalchemy_utils.types.uuid.UUIDType()),
+  column('item_collection_id', sqlalchemy_utils.types.uuid.UUIDType()),
+  column('seq', sa.Integer()),
+  column('created_at', sa.DateTime()))
+
+
+def upgrade():
+    op.add_column('item', sa.Column('seq', sa.Integer(), nullable=True))
+    connection = op.get_bind()
+    item_collections = connection.execute(sa.select([item_collection.c.id]))
```
jace: Shouldn't this select specify a sort order based on id or created_at?

boxoffice/models/category.py
```diff
@@ -15,3 +15,7 @@ class Category(BaseScopedNameMixin, db.Model):
     item_collection = db.relationship(ItemCollection, backref=db.backref('categories', cascade='all, delete-orphan'))
 
     parent = db.synonym('item_collection')
+
+    @classmethod
+    def get_by_item_collection(cls, item_collection):
```
jace: Make this a relationship in ItemCollection. This method is unnecessary.

jace: Use ordering_list for categories as well.

PR #80, opened 2016-07-22
Adds support for signed discount coupon codes.

shreyas-satish:


alembic/versions/510d2db2b008_add_signed_code_support.py
```diff
@@ -0,0 +1,24 @@
...
+revision = '510d2db2b008'
+down_revision = '74770336785'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.alter_column('discount_coupon', 'code', existing_type=sa.VARCHAR(length=20), type_=sa.VARCHAR(length=50))
+    op.create_unique_constraint('discount_policy_discount_code_base_key', 'discount_policy', ['discount_code_base'])
```
jace: What's `discount_code_base`? A previous column?

boxoffice/models/discount_policy.py
```diff
@@ -43,6 +45,9 @@ class DiscountPolicy(BaseScopedNameMixin, db.Model):
     percentage = db.Column(db.Integer, nullable=True)
     # price-based discount
     is_price_based = db.Column(db.Boolean, default=False, nullable=False)
+    discount_code_base = db.Column(db.Unicode(20), nullable=True)
```
jace: This column is missing in the migration.

alembic/versions/510d2db2b008_add_signed_code_support.py
shreyas-satish: Yes. It was added with signed codes in mind.

boxoffice/models/discount_policy.py
```diff
@@ -53,6 +58,33 @@ def is_automatic(self):
     def is_coupon(self):
         return self.discount_type == DISCOUNT_TYPE.COUPON
 
+    def gen_signed_code(self):
+        """Generates a signed code in the format discount_code_base.randint.signature"""
+        signer = Signer(self.secret)
+        key = "{base}.{randint}".format(base=self.discount_code_base, randint=random.randint(1, 10000))
```
jace: Don't hardcode this for randint. Put that in the view. We'll have two scenarios:

1. A company wants a few thousand codes, so we make them sequence numbers 1..1000. The policy is linked to their account (once Lastuser integration is in place), so they can login to the dashboard and copy codes over or re-download them.

2. We're mailing our database with a referral code. As the sequence numbers here give away the size of the database, or your position number, we use randint. However, because it's randint, there's no way to track who was given what number, or to retrieve that number. These codes can be generated once only. There's no other way to ensure two different people weren't given the same code. Finally, because you can't use the same randint twice even within one batched call, the caller needs to generate the random numbers and populate a dictionary with {randint: generated code}.

boxoffice/models/discount_policy.py
shreyas-satish: This column was added in the first [init migration](https://github.com/hasgeek/boxoffice/blob/master/alembic/versions/4ffee334e82e_init_models.py). The reference to the column in the model was removed since it wasn't of use until signed codes were implemented.

boxoffice/models/discount_policy.py
```diff
@@ -53,6 +58,33 @@ def is_automatic(self):
     def is_coupon(self):
         return self.discount_type == DISCOUNT_TYPE.COUPON
 
+    def gen_signed_code(self):
+        """Generates a signed code in the format discount_code_base.randint.signature"""
+        signer = Signer(self.secret)
```
jace: If there's a startup cost to this call, move this to a `@cached_property` method.

boxoffice/models/discount_policy.py
```diff
@@ -53,6 +58,33 @@ def is_automatic(self):
...
+        key = "{base}.{randint}".format(base=self.discount_code_base, randint=random.randint(1, 10000))
+        return signer.sign(key)
+
+    @classmethod
```
jace: This could be `@staticmethod` since it doesn't need the `cls` parameter.

boxoffice/models/discount_policy.py
jace: That's a bug. A fresh installation will not run through migrations. It'll just create schema from the code and stamp it with the last migration number. You need to remove the line from the init migration and do a manual cleanup of production.

jace: We do fresh installations sometimes to clean up past migration errors that have left production out of sync with code.

jace: This happened to us when we moved from MySQL to PostgreSQL in 2012. Because MySQL's MyISAM storage doesn't support foreign keys, the migrated db in production was missing them for over a year, and because our form validators weren't thorough, we had data corruption until we looked at the schema and noticed it was out of sync (in 2014). We had to dump and re-import data and manually clean up conflicts to fix it.

boxoffice/models/discount_policy.py
```diff
@@ -53,6 +58,33 @@ def is_automatic(self):
...
+    def gen_signed_code(self):
+        """Generates a signed code in the format discount_code_base.randint.signature"""
+        signer = Signer(self.secret)
+        key = "{base}.{randint}".format(base=self.discount_code_base, randint=random.randint(1, 10000))
+        return signer.sign(key)
+
+    @staticmethod
+    def is_signed_code_format(code):
+        """Checks if the code is in the {x.y.z} format"""
+        return len(code.split('.')) == 3
```
jace: You'll need a validator for regular discount codes to prevent use of periods.

boxoffice/models/line_item.py
```diff
@@ -215,7 +216,17 @@ def get_from_item(cls, item, qty, coupon_codes=[]):
         coupon_policies = item.discount_policies.filter(DiscountPolicy.discount_type == DISCOUNT_TYPE.COUPON).all()
         coupon_policy_ids = [cp.id for cp in coupon_policies]
         for coupon_code in coupon_codes:
-            coupon = DiscountCoupon.query.filter(DiscountCoupon.discount_policy_id.in_(coupon_policy_ids), DiscountCoupon.code == coupon_code).one_or_none()
+            if DiscountPolicy.is_signed_code_format(coupon_code):
+                policy = DiscountPolicy.get_from_signed_code(coupon_code)
+                if policy and policy.id in coupon_policy_ids:
+                    coupon = DiscountCoupon.query.filter_by(discount_policy=policy, code=coupon_code).one_or_none()
+                    if not coupon:
+                        coupon = DiscountCoupon(id=uuid.uuid4(), discount_policy=policy, code=coupon_code, usage_limit=1, used_count=0)
```
jace: Use `uuid1mc` not `uuid4`. This should be in DiscountCoupon's constructor, not here.

tests/fixtures.py
```diff
@@ -141,3 +142,14 @@ def init_data():
...
         db.session.add(zero_discount_price)
         db.session.commit()
+
+        signed = DiscountPolicy(title='Signed',
+            item_quantity_min=1,
+            discount_type=DISCOUNT_TYPE.COUPON,
+            percentage=10,
+            discount_code_base='signed',
+            secret=uuid.uuid4().hex,
+            organization=rootconf)
```
jace: This should perhaps be a classmethod that constructs signed discount policies, like `DiscountPolicy.new_bulk()` so that `secret` isn't left to the caller. Also, "bulk" is a more user friendly term than "signed".

boxoffice/models/discount_policy.py
shreyas-satish: > There's no other way to ensure two different people weren't given the same code.
Right. What if we use `buid` here instead of `randint`?

boxoffice/models/discount_policy.py
```diff
@@ -43,6 +45,12 @@ class DiscountPolicy(BaseScopedNameMixin, db.Model):
     percentage = db.Column(db.Integer, nullable=True)
     # price-based discount
     is_price_based = db.Column(db.Boolean, default=False, nullable=False)
+
+    # Note: discount_code_base was added in the initial migration,
+    # but was removed from the model for a period of time
```
jace: This comment should be in the migration to explain why the column is not being added there.

shreyas-satish: Done.

boxoffice/models/discount_policy.py
```diff
@@ -53,6 +59,42 @@ def is_automatic(self):
...
+        """Checks if the code is in the {x.y.z} format"""
+        return len(code.split('.')) == 3
+
+    @classmethod
+    def get_from_signed_code(cls, code):
+        """Returns a discount policy given a valid signed code, returns None otherwise"""
+        if not cls.is_signed_code_format(code):
+            return None
+        discount_code_base = code.split('.')[0]
+        policy = DiscountPolicy.query.filter_by(discount_code_base=discount_code_base).one_or_none()
```
jace: Shouldn't this refer to `cls`?

tests/fixtures.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+import uuid
```
jace: Unused import?

PR #88, opened 2016-08-10
added AGPL

shreyas-satish:


jace:
BoxOffice should be Boxoffice.

PR #92, opened 2016-08-22
Fix refund policy text on boxoffice widget.

vidya-ram:
Display organization's refund policy on boxoffice widget.

### DIFF
--- a/boxoffice/views/item_collection.py
+++ b/boxoffice/views/item_collection.py
@@ -64,4 +64,4 @@ def item_collection(item_collection):
         category_json = jsonify_category(category)
         if category_json:
             categories_json.append(category_json)
-    return jsonify(html=render_template('boxoffice.html'), categories=categories_json)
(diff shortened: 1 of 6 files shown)


boxoffice/views/item_collection.py
```diff
@@ -64,4 +64,4 @@ def item_collection(item_collection):
         category_json = jsonify_category(category)
         if category_json:
             categories_json.append(category_json)
-    return jsonify(html=render_template('boxoffice.html'), categories=categories_json)
+    return jsonify(html=render_template('boxoffice.html'), categories=categories_json, refund_policy=item_collection.organization.details.get('refund_policy'))
```
jace: This needs a default value.

tests/test_item_collection.py
```diff
@@ -7,7 +7,7 @@
 
 class TestItemCollectionAPI(unittest.TestCase):
 
-    expected_keys = ["categories", "html"]
+    expected_keys = ["categories", "html", "refund_policy"]
```
jace: These strings should be in single quotes as they are not UI strings. Same below.

Issue #61, opened 2016-06-30
Check for current_price before line item calculation for items with expired prices


shreyas-satish:
A request may be made for an item whose price has expired, hence resulting in `current_price` being `None` in https://github.com/hasgeek/boxoffice/blob/master/boxoffice/models/line_item.py#L95. The `current_price` must be checked, prior to calculating the line_item's amounts.

shreyas-satish:
Also, check whether the item exists [here](https://github.com/hasgeek/boxoffice/blob/master/boxoffice/models/line_item.py#L92).

PR #97, opened 2016-09-02
added checks for items with expired prices

shreyas-satish:
Fixes https://github.com/hasgeek/boxoffice/issues/61

### DIFF
--- a/boxoffice/models/item.py
+++ b/boxoffice/models/item.py
@@ -64,6 +64,10 @@ class Item(BaseScopedNameMixin, db.Model):
     def quantity_available(self):
         return self.quantity_total - self.get_confirmed_line_items.count()
 
+    def is_available(self):
+        """Checks if an item has a current price and has a positive quantity_available"""
+        return bool(self.current_price() and self.quantity_available > 0)
+
 
 class Price(BaseScopedNameMixin, db.Model):
     __tablename__ = 'price'
--- a/boxoffice/models/item.py
+++ b/boxoffice/models/item.py
@@ -64,6 +64,7 @@ class Item(BaseScopedNameMixin, db.Model):
     def quantity_available(self):
         return self.quantity_total - self.get_confirmed_line_items.count()
 
+    @property
     def is_available(self):
         """Checks if an item has a current price and has a positive quantity_available"""
         return bool(self.current_price() and self.quantity_available > 0)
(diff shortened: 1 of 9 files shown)


boxoffice/models/item.py
```diff
@@ -64,6 +64,11 @@ def get_by_category(cls, category):
     def quantity_available(self):
         return self.quantity_total - self.get_confirmed_line_items.count()
 
+    @property
+    def is_available(self):
+        """Checks if an item has a current price and has a positive quantity_available"""
+        return bool(self.current_price() and self.quantity_available > 0)
```
jace: What if `current _price()` is `0` because it is a free item? Do an `is not None` check.

shreyas-satish: `current_price()` can't be `0`. In case of a free item, `current_price()` will return a `Price` object, whose `amount` will be `0`.

boxoffice/models/line_item.py
```diff
@@ -96,17 +96,23 @@ def calculate(cls, line_item_dicts, coupons=[]):
...
+                base_amount=item.current_price().amount if item.is_available else None))
+
         for item_id in item_line_items.keys():
-            item_line_items[item_id] = discounter.get_discounted_line_items(item_line_items[item_id], coupon_list)
-            line_items.extend(item_line_items[item_id])
+            if item_line_items[item_id]['is_available']:
+                line_items.extend(discounter.get_discounted_line_items(item_line_items[item_id]['line_items'], coupon_list))
+            else:
+                # item isn't available, no need to calculate discounts
+                line_items.extend(item_line_items[item_id]['line_items'])
```
jace: I don't get what is happening here. Two lines above, if the item is not available, you add `None` to the list. And now if the value is falsy, you're doing something with it. What exactly?

shreyas-satish: Simplified the flow and added comments - https://github.com/hasgeek/boxoffice/pull/97/files#diff-5e9e8b8df4de8db454de121fd430de77R112

shreyas-satish: @jace Is it better now?

boxoffice/models/item.py
jace: Perhaps a comment clarifying that the `current_price()` call returns a `Price` object?

shreyas-satish: Done.

PR #100, opened 2016-09-06
add utm_campaign tracking in order

shreyas-satish:


boxoffice/forms/order.py
```diff
@@ -35,3 +35,7 @@ class BuyerForm(forms.Form):
     email = forms.EmailField(__("Email"), validators=[forms.validators.DataRequired(), forms.validators.Length(max=80)])
     fullname = forms.StringField(__("Full name"), validators=[forms.validators.DataRequired()])
     phone = forms.StringField(__("Phone number"), validators=[forms.validators.Length(max=16)])
+
+
+class OrderForm(forms.Form):
+    utm_campaign = forms.StringField(__("UTM Campaign"), validators=[forms.validators.Length(max=80)])
```
jace: You can't fail validation as there's no user input. Do what Hasjob does instead, capturing whatever is in the request and letting the database validate if the capture had an error.

boxoffice/models/order.py
```diff
@@ -114,3 +114,26 @@ def is_fully_assigned(self):
...
+
+    # Google Analytics parameters
+    utm_source = db.Column(db.Unicode(250), nullable=False, default=u'')
+    utm_medium = db.Column(db.Unicode(250), nullable=False, default=u'')
+    utm_term = db.Column(db.Unicode(250), nullable=False, default=u'')
+    utm_content = db.Column(db.Unicode(250), nullable=False, default=u'')
+    utm_id = db.Column(db.Unicode(250), nullable=False, default=u'')
+    utm_campaign = db.Column(db.Unicode(250), nullable=False, default=u'')
+    # Google click id (for AdWords)
+    gclid = db.Column(db.Unicode(250), nullable=False, default=u'')
```
jace: You'll need `index=True` for all/most columns as you'll want to query by these values.

jace: I suspect there will also be performance issues despite the index, as you'll want results scoped by orders that refer to an item collection, so it's a significant join. Let's deal with that later.

boxoffice/views/order.py
```diff
@@ -149,13 +150,20 @@ def order(item_collection):
...
+    order_session_form = OrderSessionForm.from_json(request.json.get('order_session'))
     order = Order(user=user,
         organization=item_collection.organization,
         item_collection=item_collection,
         buyer_email=buyer_form.email.data,
         buyer_fullname=buyer_form.fullname.data,
         buyer_phone=buyer_form.phone.data)
 
+    order_session = OrderSession(**order_session_form.data)
+    if order_session.utm_campaign:
```
jace: Is there a context in which you'll get other fields but not `utm_campaign`? What does Sendgrid include in links? (Google Analytics documentation should be able to clarify this.)

boxoffice/forms/order.py
```diff
@@ -35,3 +35,13 @@ class BuyerForm(forms.Form):
...
+
+
+class OrderSessionForm(forms.Form):
+    utm_campaign = forms.StringField(__("UTM Campaign"))
+    utm_source = forms.StringField(__("UTM Source"))
+    utm_medium = forms.StringField(__("UTM Medium"))
+    utm_term = forms.StringField(__("UTM Term"))
+    utm_content = forms.StringField(__("UTM Content"))
+    utm_id = forms.StringField(__("UTM Id"))
+    gclid = forms.StringField(__("Gclid"))
```
jace: The form needs a `validate` method that ensures all fields are within the length limit and raises `ValidationError` if a required field (like say `utm_campaign`) is missing.

boxoffice/views/order.py
jace: Change these two lines:

```python
order_session = OrderSession(**order_session_form.data)
if order_session.utm_campaign:
```

to:

```python
if order_session_form.validate():
    order_session = OrderSession()
    order_session_form.populate_obj(order_session)
```

boxoffice/views/order.py
```diff
@@ -149,13 +150,20 @@ def order(item_collection):
...
+        order_session.order = order
+        order_session.referrer = unicode_http_header(request.referrer)[:2083] if request.referrer else None
```
jace: Referrer should go through the form as well.

boxoffice/forms/order.py
shreyas-satish: > You can't fail validation as there's no user input.

@jace I'm a little confused because I was under the impression that the validation was supposed to be pushed down to the database, because of the above comment.

boxoffice/forms/order.py
```diff
@@ -2,9 +2,10 @@
 
 from baseframe import __
 import baseframe.forms as forms
+from baseframe.forms import ValidationError
```
jace: You're importing but not raising ValidationError.

boxoffice/models/order.py
jace: Index is still pending.

PR #101, opened 2016-09-12
Admin page to download reports.

vidya-ram:
First version: Option to download list of attendees in csv format.

screen shot 2016-09-13 at 10 43 03 am


boxoffice/views/admin_report.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+@app.route('/admin/ic/<ic_id>/reports/order')
+@lastuser.requires_login
+@load_models(
+    (ItemCollection, {'id': 'ic_id'}, 'item_collection'),
+    permission='org_admin'
+    )
+def admin_order_report(item_collection):
+    orders = Order.query.filter_by(item_collection=item_collection).order_by('created_at').all()
```
shreyas-satish: This is going to fetch purchase orders as well. If we only care about sales orders and invoiced orders, then:

```python
    orders = item_collection.orders.filter(Order.status.in_(ORDER_STATUS.CONFIRMED)).order_by('created_at')
```

boxoffice/views/admin_report.py
```diff
@@ -0,0 +1,102 @@
...
+    outfile = StringIO()
+    out = unicodecsv.writer(outfile, encoding='utf-8')
+    out.writerow(order_headers)
+    for order in orders:
+        if (order.is_confirmed):
```
shreyas-satish: This won't be required with the new query.

boxoffice/views/admin_report.py
```diff
@@ -0,0 +1,102 @@
...
+
+import unicodecsv
+from cStringIO import StringIO
+from flask import jsonify, Response
+from .. import app, lastuser
+from coaster.views import load_models, render_with
+from boxoffice.models import ItemCollection, Order
+
+
+order_headers = [
```
shreyas-satish: Why is this outside the function in which it is used in?

boxoffice/views/admin_report.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def jsonify_report(data_dict):
+    return jsonify(org_name=data_dict['item_collection'].organization.name, title=data_dict['item_collection'].title)
+
+
+@app.route('/admin/ic/<ic_id>/reports')
+@lastuser.requires_login
+@load_models(
+    (ItemCollection, {'id': 'ic_id'}, 'item_collection'),
```
jace: You should be using `id` instead of `ic_id` here and in the route above.

boxoffice/views/admin_report.py
```diff
@@ -0,0 +1,44 @@
...
+    permission='org_admin'
+    )
+@render_with({'text/html': 'index.html', 'application/json': jsonify_report}, json=True)
```
jace: Don't use `json=True` if you're providing an explicit `application/json` handler. It provides another for `text/json` which can be a backdoor into your code.

boxoffice/views/admin_report.py
```diff
@@ -0,0 +1,44 @@
...
+def admin_report(item_collection):
+    return dict(title=item_collection.organization.title, item_collection=item_collection)
+
+
+def generate_csv(headers, rows):
+    for row in headers + rows:
+        yield ','.join([unicode(attr) for attr in row]) + '\n'
```
jace: Use the `unicodecsv` library for this. Don't handroll CSV.

boxoffice/views/admin_report.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@app.route('/admin/ic/<ic_id>/reports/tickets.csv')
+@lastuser.requires_login
+@load_models(
+    (ItemCollection, {'id': 'ic_id'}, 'item_collection'),
```
jace: `id` as above.

boxoffice/views/admin_report.py
```diff
@@ -0,0 +1,44 @@
...
+    permission='org_admin'
+    )
+def admin_line_items_report(item_collection):
+    line_item_join = db.outerjoin(LineItem, Assignee).outerjoin(DiscountPolicy).join(Item).join(Order)
```
jace: Can you confirm what `outerjoin` vs a regular `join` does?

boxoffice/views/admin_report.py
```diff
@@ -0,0 +1,44 @@
...
+    line_item_query = db.select([LineItem.id, Item.title, LineItem.base_amount, LineItem.discounted_amount, LineItem.final_amount, DiscountPolicy.title, Order.buyer_fullname, Order.buyer_email, Order.buyer_phone, Assignee.fullname, Assignee.email, Assignee.phone, Assignee.details])\
+        .select_from(line_item_join)\
+        .where(LineItem.status == LINE_ITEM_STATUS.CONFIRMED).where(Assignee.current == True).where(Order.item_collection == item_collection)\
+        .order_by('created_at')
```
jace: Use of `\` will trigger PyFlakes errors if it's in use (as it should be).

boxoffice/views/admin_report.py
shreyas-satish: This is because we also want to fetch line items that have not been assigned yet and line items that are not associated with any discounts.

jace: A comment in the code will help.

boxoffice/views/admin_report.py
```diff
@@ -0,0 +1,35 @@
...
+
+from flask import Response, jsonify
+from .. import app, lastuser
+from coaster.views import load_models, render_with
+from boxoffice.models import db, ItemCollection, Order, Item, LineItem, Assignee, DiscountPolicy, LINE_ITEM_STATUS
+from boxoffice.views.utils import csv_response
+
+
+def jsonify_report(data_dict):
+    return jsonify(org_name=data_dict['item_collection'].organization.name, title=data_dict['item_collection'].title)
```
jace: The `title` you're returning in the JSON version is different from the HTML version?

boxoffice/views/utils.py
```diff
@@ -69,3 +71,16 @@ def wrapper(*args, **kwargs):
...
+    """
+    Returns a response, with mimetype set to text/csv,
+    given a list of headers and a two-dimensional list of rows
+    """
+    stream = StringIO()
+    csv_writer = unicodecsv.writer(stream)
+    csv_writer.writerow(headers)
+    for row in rows:
+        csv_writer.writerow(row)
+    return Response(stream.getvalue(), mimetype='text/csv')
```
shreyas-satish: It doesn't display inline as of now, the browser prompts a download.

PR #106, opened 2016-09-21
Transferrable until

vidya-ram:
1. Add ability to block transfers after a certain date.
2. CC past assignee incase of ticket transfer.
3. Changed variable names from attendee to assignee.


boxoffice/views/participant.py
```diff
@@ -38,11 +39,16 @@ def assign(order):
         db.session.commit()
     else:
         if line_item.current_assignee:
+            past_assignee = line_item.current_assignee.email
```
shreyas-satish: I suggest `previous_assignee_email` instead of `past_assignee`.

boxoffice/mailclient.py
```diff
@@ -55,15 +55,19 @@ def send_line_item_cancellation_mail(line_item_id, subject="Ticket Cancellation"
 
 
 @job('boxoffice')
-def send_ticket_assignment_mail(line_item_id):
+def send_ticket_assignment_mail(line_item_id, past_assignee):
```
shreyas-satish: This method can be litte more lightweight. It can just accept a list of recipients from the caller, instead of having knowledge of a 'past assignee'.

boxoffice/mailclient.py
```diff
@@ -55,15 +55,19 @@ def send_line_item_cancellation_mail(line_item_id, subject="Ticket Cancellation"
...
     """
     Sends a confirmation email once details are filled and ticket has been assigned.
     """
     with app.test_request_context():
         line_item = LineItem.query.get(line_item_id)
         order = line_item.order
         subject = order.item_collection.title + ": Here's your ticket"
-        msg = Message(subject=subject, recipients=[line_item.current_assignee.email], bcc=[order.buyer_email])
+        if past_assignee:
+          recipient_list = [order.buyer_email, past_assignee]
```
shreyas-satish: This should be `cc_list` since recipient would imply the intended recipient of the email.

boxoffice/views/participant.py
vidya-ram: Done

boxoffice/mailclient.py
vidya-ram: Updated

boxoffice/mailclient.py
vidya-ram: Done

boxoffice/mailclient.py
```diff
@@ -55,15 +55,15 @@ def send_line_item_cancellation_mail(line_item_id, subject="Ticket Cancellation"
 
 
 @job('boxoffice')
-def send_ticket_assignment_mail(line_item_id):
+def send_ticket_assignment_mail(line_item_id, recipient_list, cc_list):
```
shreyas-satish: I think `cc_list` should have `[]` as its default parameter since it isn't mandatory.

PR #110, opened 2016-10-13
Pagination for discount policies

iambibhas:
This adds a pagination bar at the bottom.

screenshot_20161013_102612


boxoffice/views/admin_discount.py
```diff
@@ -51,11 +65,35 @@ def jsonify_discount_policies(data_dict):
 @render_with({'text/html': 'index.html', 'application/json': jsonify_discount_policies}, json=True)
 def admin_discount_policies(organization):
     query = request.args.get('search')
-    if query:
-        discount_policies = DiscountPolicy.query.filter(DiscountPolicy.title.ilike('%{query}%'.format(query=query))).all()
+    page = request.args.get('page', 1)
+
+    if request.is_xhr:
+        if query:
+            discount_policies = DiscountPolicy.query.filter(DiscountPolicy.title.ilike('%{query}%'.format(query=query)))
```
shreyas-satish: For instance, if I search for `JSFoo`, this might theoretically return 20 results, which in turn has to be paginated.

boxoffice/views/admin_discount.py
```diff
@@ -51,11 +65,35 @@ def jsonify_discount_policies(data_dict):
...
+        else:
+            discount_policies = DiscountPolicy.query.filter(DiscountPolicy.organization == organization).order_by('created_at desc')
+
+        total_policies = discount_policies.count()
+
+        try:
+            page = int(page)
+            offset = (page - 1) * PAGE_SIZE
+            discount_policies = discount_policies.limit(PAGE_SIZE).offset(offset)
+        except ValueError:
```
shreyas-satish: What's a case in which this would be raised?

boxoffice/views/admin_discount.py
```diff
@@ -9,13 +10,20 @@
 from utils import xhr_only, date_time_format
 
 
+# Move this to settings maybe
+PAGE_SIZE = 6
```
shreyas-satish: This can just be a local variable inside `admin_discount_policies`?

boxoffice/views/admin_discount.py
```diff
@@ -51,11 +65,35 @@ def jsonify_discount_policies(data_dict):
 @render_with({'text/html': 'index.html', 'application/json': jsonify_discount_policies}, json=True)
 def admin_discount_policies(organization):
     query = request.args.get('search')
```
shreyas-satish: The request argument can just be `q`?

boxoffice/views/admin_discount.py
```diff
@@ -51,11 +65,35 @@ def jsonify_discount_policies(data_dict):
...
+        total_policies = discount_policies.count()
+
+        try:
+            page = int(page)
+            offset = (page - 1) * PAGE_SIZE
+            discount_policies = discount_policies.limit(PAGE_SIZE).offset(offset)
+        except ValueError:
+            pass
+
+        print page
```
shreyas-satish: Remove `print`.

boxoffice/views/admin_discount.py
```diff
@@ -51,11 +65,35 @@ def jsonify_discount_policies(data_dict):
...
+            discount_policies = DiscountPolicy.query.filter(DiscountPolicy.title.ilike('%{query}%'.format(query=query)))
+        else:
+            discount_policies = DiscountPolicy.query.filter(DiscountPolicy.organization == organization).order_by('created_at desc')
+
+        total_policies = discount_policies.count()
+
+        try:
+            page = int(page)
+            offset = (page - 1) * PAGE_SIZE
+            discount_policies = discount_policies.limit(PAGE_SIZE).offset(offset)
```
shreyas-satish: Can we rename `PAGE_SIZE` to `posts_per_page`?

boxoffice/views/admin_discount.py
```diff
@@ -51,11 +65,35 @@ def jsonify_discount_policies(data_dict):
 @render_with({'text/html': 'index.html', 'application/json': jsonify_discount_policies}, json=True)
 def admin_discount_policies(organization):
     query = request.args.get('search')
-    if query:
-        discount_policies = DiscountPolicy.query.filter(DiscountPolicy.title.ilike('%{query}%'.format(query=query))).all()
+    page = request.args.get('page', 1)
```
shreyas-satish: Maybe convert it to an `int` right here?

boxoffice/views/admin_discount.py
```diff
@@ -51,11 +65,35 @@ def jsonify_discount_policies(data_dict):
...
+
+    if request.is_xhr:
```
shreyas-satish: Why is this required?

boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -25,6 +25,10 @@ export const DiscountPolicyView = {
           org: org_name,
           title: title,
           discount_policies: discount_policies,
+          paginated: paginated,
+          total_pages: total_pages,
+          current_page: current_page,
+          pages: Array.from(Array(total_pages).keys()).map(x => ++x),
```
iambibhas: Making an array starting from `1` to `total_pages`. It's an one liner. But now I feel a simple for loop will be more readable. Reactive doesn't seem to have a way to run a loop for a given length, so have to send a predefined array.

boxoffice/views/admin_discount.py
iambibhas: I was thinking if we should make it a global config. so we just need to change it in one place. defining it inside the view also works for now.

boxoffice/views/admin_discount.py
iambibhas: Sure.

boxoffice/views/admin_discount.py
iambibhas: @shreyas-satish the actual data is loaded when the ajax call is made. when you load the page, it loads the JS that calls the endpoint again to load data. so if we dont have this here, the same db queries will be called twice but only the ajax call will update the page.

iambibhas: @shreyas-satish the actual data is loaded when the ajax call is made. when you load the page, it loads the JS that calls the endpoint again to load data. so if we dont have this here, the same db queries will be called twice but only the ajax call will update the page.

boxoffice/views/admin_discount.py
```diff
@@ -52,12 +62,38 @@ def jsonify_discount_policies(data_dict):
     )
 @render_with({'text/html': 'index.html', 'application/json': jsonify_discount_policies}, json=True)
 def admin_discount_policies(organization):
+    RESULTS_PER_PAGE = 6
```
shreyas-satish: Why only 6?

boxoffice/views/admin_discount.py
```diff
@@ -52,12 +62,38 @@ def jsonify_discount_policies(data_dict):
...
     query = request.args.get('search')
```
shreyas-satish: I suggest renaming this to `search_query`. It's easy to confuse this with SQLAlchemy's `query.`

boxoffice/views/admin_discount.py
```diff
@@ -52,12 +62,38 @@ def jsonify_discount_policies(data_dict):
...
+                DiscountPolicy.title.ilike('%{query}%'.format(query=query))
+            )
+
+        total_policies = discount_policies.count()
+
+        offset = (page - 1) * RESULTS_PER_PAGE
+        discount_policies = discount_policies.limit(RESULTS_PER_PAGE).offset(offset)
+
+        return dict(
+            org=organization, title=organization.title,
```
shreyas-satish: Why are we passing both `organization` and `organization.title`?

boxoffice/views/admin_discount.py
```diff
@@ -52,12 +62,38 @@ def jsonify_discount_policies(data_dict):
...
+            discount_policies=discount_policies,
+            total_pages=int(math.ceil(total_policies/RESULTS_PER_PAGE)),
+            current_page=page
+        )
     else:
-        discount_policies = DiscountPolicy.query.filter(DiscountPolicy.organization == organization).order_by('created_at desc').all()
-    return dict(org=organization, title=organization.title, discount_policies=discount_policies)
+        return dict(
+            org=organization, title=organization.title
```
shreyas-satish: Same as above.

boxoffice/views/admin_discount.py
```diff
@@ -52,12 +62,38 @@ def jsonify_discount_policies(data_dict):
...
-        discount_policies = DiscountPolicy.query.filter(DiscountPolicy.title.ilike('%{query}%'.format(query=query))).all()
+    try:
+        page = int(request.args.get('page', 1))
+    except ValueError:
+        page = 1
+
+    if request.is_xhr:
+        discount_policies = DiscountPolicy.query.filter(
+            DiscountPolicy.organization == organization
+        ).order_by('created_at desc')
```
shreyas-satish: Perhaps better to set this as the default sort order and use `organization.discount_policies`.

boxoffice/views/admin_discount.py
```diff
@@ -52,12 +62,38 @@ def jsonify_discount_policies(data_dict):
     )
 @render_with({'text/html': 'index.html', 'application/json': jsonify_discount_policies}, json=True)
 def admin_discount_policies(organization):
+    RESULTS_PER_PAGE = 6
     query = request.args.get('search')
-    if query:
-        discount_policies = DiscountPolicy.query.filter(DiscountPolicy.title.ilike('%{query}%'.format(query=query))).all()
+    try:
+        page = int(request.args.get('page', 1))
+    except ValueError:
```
shreyas-satish: This perhaps can just be an `if/else`?

boxoffice/views/admin_discount.py
iambibhas: Was testing with a low number to test pagination. what's a good number per page? given I only see 3 items per row.

boxoffice/views/admin_discount.py
iambibhas: This is casting the query param to integer. What do you mean by using `if/else`?

shreyas-satish: I meant if we could use an `if/else` instead of using a `try/catch`. i.e

```python
if request.args.get('page'):
    page = int(request.args.get('page'))
else:
    page = 1
```

boxoffice/views/admin_discount.py
shreyas-satish: 20?

boxoffice/views/admin_discount.py
iambibhas: @shreyas-satish ah. ok. I'm handling both the case of `page` being present and the chance that it might not be a number. In the simple if/else, it'll throw exception if page is not integer.

shreyas-satish: Fair enough.

boxoffice/views/admin_discount.py
jace: Use `coaster.views.requestargs`. Don't read parameters directly from `request.args`

boxoffice/views/admin_discount.py
vidya-ram: For this https://github.com/hasgeek/boxoffice/blob/master/boxoffice/templates/index.html#L7

For different view, organization might not be present to use organization.title in the template
https://github.com/hasgeek/boxoffice/blob/master/boxoffice/views/admin_report.py#L22

boxoffice/views/admin_discount.py
iambibhas: @jace any example on how to use `requestargs` on a view along with `load_models()`?

iambibhas: nvm. found how to use it. thanks.

PR #113, opened 2016-10-25
add usage_limit to discount policy for bulk coupons

shreyas-satish:
This change will allow coupons generated in bulk, which aren't stored during generation, to have a customizable usage limit.

### DIFF
--- a/boxoffice/models/discount_policy.py
+++ b/boxoffice/models/discount_policy.py
@@ -50,6 +50,9 @@ class DiscountPolicy(BaseScopedNameMixin, db.Model):
     secret = db.Column(db.Unicode(50), nullable=True)
 
     items = db.relationship('Item', secondary=item_discount_policy)
+    # Coupons generated in bulk are not stored in the database during generation.
+    # This field allows specifying the number of times a coupon, generated in bulk, can be used
+    bulk_coupon_usage_limit = db.Column(db.Integer, nullable=True, default=1)
 
     @cached_property
     def is_automatic(self):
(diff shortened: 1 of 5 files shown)


boxoffice/models/discount_policy.py
```diff
@@ -50,6 +50,9 @@ class DiscountPolicy(BaseScopedNameMixin, db.Model):
     secret = db.Column(db.Unicode(50), nullable=True)
 
     items = db.relationship('Item', secondary=item_discount_policy)
+    # Coupons generated in bulk are not stored in the database during generation.
+    # This field allows specifying the number of times a coupon, generated in bulk, can be used
+    bulk_coupon_usage_limit = db.Column(db.Integer, nullable=True, default=1)
```
jace: Does this default apply to newly generated coupons or past coupons too?

boxoffice/models/line_item.py
```diff
@@ -253,7 +253,7 @@ def get_from_item(cls, item, qty, coupon_codes=[]):
                 if policy and policy.id in coupon_policy_ids:
                     coupon = DiscountCoupon.query.filter_by(discount_policy=policy, code=coupon_code).one_or_none()
                     if not coupon:
-                        coupon = DiscountCoupon(discount_policy=policy, code=coupon_code, usage_limit=1, used_count=0)
+                        coupon = DiscountCoupon(discount_policy=policy, code=coupon_code, usage_limit=policy.bulk_coupon_usage_limit, used_count=0)
```
jace: What's happening here? You're simply specifying a default value in the policy that is picked up for new coupons?

shreyas-satish: This is only for signed coupons, yes. Since the coupon isn't stored (it's only stored when it's used), the usage limit for the signed coupon is stored in the policy.

boxoffice/models/discount_policy.py
shreyas-satish: It will apply to past signed coupons too as long as they're unused.

PR #126, opened 2017-01-12
fixes #21

### LINKED ISSUE #21: Price increase status on UI
Price increase status (Prices go up after ...) on UI should be based on if a higher price is present for the item.

shreyas-satish:


boxoffice/models/item.py
```diff
@@ -43,6 +43,12 @@ def current_price(self):
         """
         return self.price_at(datetime.utcnow())
 
+    def has_higher_price(self, current_price):
```
jace: Not an immediate priority, but all these queries to render a widget are expensive. This value can be cached for up to a week (in our current use), so it needs to be pushed to cache for whatever timeline is left for validity.

I suggest model-level methods should return status and valid_until, and a wrapping function in a workflow object should push this to cache for the valid_until period.

boxoffice/templates/boxoffice.html
```diff
@@ -69,7 +69,14 @@ <h1 class="category-heading">{{title}}</h1>
                           {{/}}
                         </div>
                         {{#if line_item.quantity_available > 0}}
-                          <p class="valid-upto"><i class="fa fa-fw fa-long-arrow-up"></i>Price goes up after {{line_item.price_valid_upto}}</p>
+                          <p class="valid-upto">
+                            {{#if line_item.has_higher_price}}
+                              <i class="fa fa-fw fa-long-arrow-up"></i>
+                              Price goes up after {{line_item.price_valid_upto}}
```
jace: Isn't `valid_upto` the non-inclusive endpoint, such as "midnight on Saturday"? In that case the phrasing should be "goes up at", not "goes up after".

boxoffice/templates/boxoffice.html
```diff
@@ -69,7 +69,14 @@ <h1 class="category-heading">{{title}}</h1>
...
+                            {{else}}
+                              Sale closes on {{line_item.price_valid_upto}}
```
jace: Also "at" not "on".

requirements.txt
```diff
@@ -2,7 +2,7 @@ Flask
 Flask-SQLAlchemy
 Flask-Assets
 Flask-Mail
-Flask-WTF
+Flask-WTF==0.12
```
jace: What broke?

boxoffice/templates/boxoffice.html
shreyas-satish: `line_item.price_valid_upto` is formatted as a date, eg: Jan 16 2017. Okay to stick with "goes up after".

boxoffice/templates/boxoffice.html
shreyas-satish: `price_valid_upto` is formatted as a date. "on" works?

requirements.txt
shreyas-satish: The latest [version](https://flask-wtf.readthedocs.io/en/stable/changelog.html), seems to have changed the API to selectively disable CSRF token checks.

jace: Let's do a site-wide upgrade then. Can't risk getting bogged down with an old release.

boxoffice/templates/boxoffice.html
jace: "on" works then. You'll need a better way to deal with such nuances though. I don't see any formatting options here in the template, where you'd expect to choose the formatting.

boxoffice/templates/boxoffice.html
jace: Isn't that still "goes up on"? We shouldn't be picking 11:59 PM as end-point. It should be a non-inclusive endpoint at midnight.

PR #127, opened 2017-01-16
coerce coupon values into unicode

shreyas-satish:


boxoffice/views/order.py
```diff
@@ -77,8 +77,8 @@ def jsonify_order(data):
 def sanitize_coupons(coupons):
     if type(coupons) is not list:
```
jace: This should be `if not isinstance(coupons, list):`. Use of `type` was deprecated in Python 2.2.

PR #130, opened 2017-01-27
JSON API for weekly sales numbers

shreyas-satish:


boxoffice/views/admin.py
```diff
@@ -35,3 +38,16 @@ def jsonify_org(data):
...
+
+@app.route('/admin/o/<org>/revenue')
+@load_models(
+    (Organization, {'name': 'org'}, 'organization')
+    )
+def org_revenue(organization):
+    if not request.args.get('access_token') or request.args.get('access_token') != organization.details.get("access_token"):
+        abort(401)
+    item_collection_ids = [item_collection.id for item_collection in organization.item_collections]
+    timezone = g.user.timezone if g.user else app.config.get('TIMEZONE')
```
jace: You can't be using `g.user.timezone` for an API call.

jace: I'd say you have a problem here that needs resolution. The timezone you're returning data for needs to be consistent. It should be a call parameter.

boxoffice/views/admin.py
```diff
@@ -35,3 +38,16 @@ def jsonify_org(data):
...
+    year = int(request.args.get('year')) if request.args.get('year') else localize_timezone(datetime.datetime.utcnow(), tz=timezone).year
```
jace: Don't assume current year. Make this a mandatory argument as well.

boxoffice/views/admin.py
```diff
@@ -35,3 +38,16 @@ def jsonify_org(data):
...
+    return jsonify(weekly_sales=calculate_weekly_sales(item_collection_ids, timezone, year))
```
jace: Make a consistent API call format. We agreed on `{"status": "ok", "result": …}`

boxoffice/views/admin.py
```diff
@@ -35,3 +38,16 @@ def jsonify_org(data):
 @render_with({'text/html': 'index.html', 'application/json': jsonify_org}, json=True)
 def org(organization):
     return dict(org=organization, title=organization.title)
+
+
+@app.route('/admin/o/<org>/revenue')
```
jace: Version number missing. Also, why is this under `/admin` and not the agreed on format of `/api/<version>/<resource>/<action>`?

boxoffice/views/admin.py
```diff
@@ -35,3 +38,16 @@ def jsonify_org(data):
...
+@load_models(
+    (Organization, {'name': 'org'}, 'organization')
+    )
+def org_revenue(organization):
+    if not request.args.get('access_token') or request.args.get('access_token') != organization.details.get("access_token"):
+        abort(401)
```
jace: Move the access token check into a decorator.

boxoffice/models/line_item.py
```diff
@@ -255,10 +256,30 @@ def sales_by_date(dates, user_tz, item_ids):
...
     """
-    Calculates the percentage difference in sales between today and yesterday
+    Calculates sales per week for items in the given set of item_collection_ids in a given year,
+    in the user's timezone.
     """
+    week_sales = db.session.query('sales_week', 'sum').from_statement('''SELECT DATE_TRUNC('week', ordered_at AT TIME ZONE 'UTC' AT TIME ZONE :timezone)
+        AS sales_week, SUM(final_amount)
+        FROM line_item INNER JOIN item on line_item.item_id = item.id
+        WHERE status=:status AND item_collection_id IN :item_collection_ids AND
+        DATE_TRUNC('year', ordered_at AT TIME ZONE 'UTC' AT TIME ZONE :timezone) = :year GROUP BY sales_week ORDER BY sales_week;
```
jace: This comparision for `year` forces it to scan every single row in the db. You need to be checking `ordered_at` >= earliest point in the year, in UTC, and < same for next year.

boxoffice/models/line_item.py
```diff
@@ -255,10 +256,30 @@ def sales_by_date(dates, user_tz, item_ids):
...
+        ''').params(timezone=user_tz, status=LINE_ITEM_STATUS.CONFIRMED, item_collection_ids=tuple(item_collection_ids), year='{year}-01-01'.format(year=year)).all()
+    week_sales_dict = {}
+    for week_sale in week_sales:
+        week_sales_dict[week_sale[0].strftime('%V')] = week_sale[1]
```
jace: `%V` is not in the documentation and will fail on Windows systems.

boxoffice/models/line_item.py
```diff
@@ -255,10 +256,30 @@ def sales_by_date(dates, user_tz, item_ids):
...
+    Calculates sales per week for items in the given set of item_collection_ids in a given year,
+    in the user's timezone.
     """
+    week_sales = db.session.query('sales_week', 'sum').from_statement('''SELECT DATE_TRUNC('week', ordered_at AT TIME ZONE 'UTC' AT TIME ZONE :timezone)
+        AS sales_week, SUM(final_amount)
+        FROM line_item INNER JOIN item on line_item.item_id = item.id
+        WHERE status=:status AND item_collection_id IN :item_collection_ids AND
+        DATE_TRUNC('year', ordered_at AT TIME ZONE 'UTC' AT TIME ZONE :timezone) = :year GROUP BY sales_week ORDER BY sales_week;
+        ''').params(timezone=user_tz, status=LINE_ITEM_STATUS.CONFIRMED, item_collection_ids=tuple(item_collection_ids), year='{year}-01-01'.format(year=year)).all()
+    week_sales_dict = {}
```
jace: Use `defaultdict(int)` to avoid the zero problem below.

boxoffice/models/line_item.py
```diff
@@ -255,10 +256,30 @@ def sales_by_date(dates, user_tz, item_ids):
...
+    for week_sale in week_sales:
+        week_sales_dict[week_sale[0].strftime('%V')] = week_sale[1]
+    for year_week in Week.weeks_of_year(year):
+        if not week_sales_dict.get(str(year_week[1])):
+            # No sales in this week, set to 0
+            week_sales_dict[str(year_week[1])] = 0
```
jace: Wait, you're settings weeks as string keys?

boxoffice/models/line_item.py
```diff
@@ -255,10 +256,30 @@ def sales_by_date(dates, user_tz, item_ids):
     return date_sales
 
 
-def sales_delta(user_tz, item_ids):
+def calculate_weekly_sales(item_collection_ids, user_tz, year):
```
jace: Why is this view helper method inside the model file?

boxoffice/models/line_item.py
```diff
@@ -255,10 +256,30 @@ def sales_by_date(dates, user_tz, item_ids):
...
+            # No sales in this week, set to 0
+            week_sales_dict[str(year_week[1])] = 0
+
+    return week_sales_dict
+
+
+def sales_delta(user_tz, item_ids):
+    """Calculates the percentage difference in sales between today and yesterday"""
     today = datetime.datetime.now().strftime("%Y-%m-%d")
     yesterday = (datetime.datetime.now() - datetime.timedelta(days=1)).strftime("%Y-%m-%d")
```
jace: Unrelated, but use `utcnow()` not `now()`

boxoffice/models/line_item.py
shreyas-satish: Since this was going to be converted to JSON (and JSON can only have string keys), the keys were stringified.

boxoffice/models/line_item.py
shreyas-satish: Since, this method runs a query against `line_item` and processes the retrieved data, it made sense to keep it inside the model, along with methods such as `sales_by_date` and `sales_delta`.

boxoffice/views/admin.py
shreyas-satish: Fixed.

boxoffice/models/line_item.py
```diff
@@ -255,12 +257,36 @@ def sales_by_date(dates, user_tz, item_ids):
...
 
 
-def sales_delta(user_tz, item_ids):
+def calculate_weekly_sales(item_collection_ids, user_tz, year):
     """
-    Calculates the percentage difference in sales between today and yesterday
+    Calculates sales per week for items in the given set of item_collection_ids in a given year,
+    in the user's timezone.
     """
+    week_sales = db.session.query('sales_week', 'sum').from_statement(text('''SELECT DATE_TRUNC('week', ordered_at AT TIME ZONE 'UTC' AT TIME ZONE :timezone)
```
jace: `text` should already be available as `db.text`.

boxoffice/models/line_item.py
```diff
@@ -255,12 +257,36 @@ def sales_by_date(dates, user_tz, item_ids):
...
+        AS sales_week, SUM(final_amount)
+        FROM line_item INNER JOIN item on line_item.item_id = item.id
+        WHERE status=:status AND item_collection_id IN :item_collection_ids
+        AND ordered_at >= :start_at AND ordered_at < :end_at
+        GROUP BY sales_week ORDER BY sales_week;
+        ''')).params(timezone=user_tz, status=LINE_ITEM_STATUS.CONFIRMED,
+        start_at='{year}-01-01'.format(year=year),
+        end_at='{year}-01-01'.format(year=year+1),
```
jace: This doesn't work. You need to include time in the desired timezone, and then cast into UTC. For example, midnight of Jan 1 in India is still Dec 31 at 6:30 PM in UTC. You're missing a few hours of data on either side without this cast.

jace: Also, string dates don't belong in here. Use datetimes.

boxoffice/models/line_item.py
```diff
@@ -255,12 +257,36 @@ def sales_by_date(dates, user_tz, item_ids):
...
+        item_collection_ids=tuple(item_collection_ids), year='{year}-01-01'.format(year=year)).all()
+    week_sales_dict = {}
+    for week_sale in week_sales:
+        week_sales_dict[int(week_sale[0].strftime('%W'))] = week_sale[1]
```
jace: `%W` isn't an ISO week number. You need `%V`, but with the caveat of when it doesn't work. Use the isoweek library to extract the week from the date.

boxoffice/models/line_item.py
jace: Also, since you're using ISO weeks, the starting point isn't necessarily Jan 1. It could be Jan 2 or 3. Find the right start for the current year.

boxoffice/models/line_item.py
```diff
@@ -255,12 +257,36 @@ def sales_by_date(dates, user_tz, item_ids):
...
+    for year_week in Week.weeks_of_year(year):
+        if not week_sales_dict.get(year_week[1]):
+            # No sales in this week, set to 0
+            week_sales_dict[year_week[1]] = 0
```
jace: Replace the `if` clause with a `setdefault` call: `week_sales_dict[year_week[1]).setdefault(0)`. Also, do this above the extraction to get the `ordereddict` benefit (unnecessary since Python 3.6 or so since everything is ordered there).

boxoffice/models/line_item.py
```diff
@@ -255,12 +257,36 @@ def sales_by_date(dates, user_tz, item_ids):
...
+        AS sales_week, SUM(final_amount)
+        FROM line_item INNER JOIN item on line_item.item_id = item.id
+        WHERE status=:status AND item_collection_id IN :item_collection_ids
+        AND ordered_at >= :start_at AND ordered_at < :end_at
+        GROUP BY sales_week ORDER BY sales_week;
+        ''')).params(timezone=user_tz, status=LINE_ITEM_STATUS.CONFIRMED,
+        start_at='{year}-01-01'.format(year=year),
+        end_at='{year}-01-01'.format(year=year+1),
+        item_collection_ids=tuple(item_collection_ids), year='{year}-01-01'.format(year=year)).all()
+    week_sales_dict = {}
```
jace: Use an ordereddict here, so when we convert to JSON, it'll be nicely ordered.

boxoffice/models/line_item.py
```diff
@@ -255,12 +257,36 @@ def sales_by_date(dates, user_tz, item_ids):
...
+        if not week_sales_dict.get(year_week[1]):
+            # No sales in this week, set to 0
+            week_sales_dict[year_week[1]] = 0
+
+    return week_sales_dict
+
+
+def sales_delta(user_tz, item_ids):
+    """Calculates the percentage difference in sales between today and yesterday"""
     today = datetime.datetime.now().strftime("%Y-%m-%d")
```
jace: This too should be `utcnow()`

boxoffice/models/line_item.py
```diff
@@ -255,12 +257,36 @@ def sales_by_date(dates, user_tz, item_ids):
...
-    yesterday = (datetime.datetime.now() - datetime.timedelta(days=1)).strftime("%Y-%m-%d")
+    yesterday = (datetime.datetime.utcnow() - datetime.timedelta(days=1)).strftime("%Y-%m-%d")
     sales = sales_by_date([today, yesterday], user_tz, item_ids)
```
jace: Your dates aren't timezone aware here. You'll have the wrong snapshot window.

boxoffice/views/admin.py
```diff
@@ -35,3 +38,26 @@ def jsonify_org(data):
 @render_with({'text/html': 'index.html', 'application/json': jsonify_org}, json=True)
 def org(organization):
     return dict(org=organization, title=organization.title)
+
+
+@app.route('/api/1/o/<org>/revenue', methods=['GET', 'OPTIONS'])
```
jace: Don't use `o` for the resource name. Use `organization`.

boxoffice/views/admin.py
```diff
@@ -35,3 +38,26 @@ def jsonify_org(data):
...
+@load_models(
+    (Organization, {'name': 'org'}, 'organization')
+    )
+def org_revenue(organization):
+    check_api_access(organization.details.get("access_token"))
```
jace: Single quotes.

boxoffice/views/admin.py
```diff
@@ -35,3 +38,26 @@ def jsonify_org(data):
...
+    if not request.args.get('year'):
+        return make_response(jsonify(status='error', message='Missing year.'), 400)
```
jace: Move this out into an `api_error` call.

boxoffice/views/admin.py
```diff
@@ -35,3 +38,26 @@ def jsonify_org(data):
...
+
+    if not request.args.get('timezone'):
+        return make_response(jsonify(status='error', message='Missing timezone.'), 400)
+
+    user_timezone = request.args.get('timezone')
+    if user_timezone not in pytz.all_timezones:
```
jace: Use `pytz.common_timezones`. `all` includes some that PostgreSQL may not recognise.

boxoffice/views/admin.py
jace: Both `api_result` and `api_success` should be in coaster.views. Move them over there.

boxoffice/models/line_item.py
shreyas-satish: Since there's no order in JSON, `jsonify` doesn't preserve order [1] even if I use `ordereddict`. I'll have to jsonify `ordereddict.items()` which will render a nested list of lists eg: [[1,100.0], [2, 200.0]]. Is that okay, or is the current structure fine?

[1] https://github.com/pallets/flask/issues/974

jace: So Flask sorts it anyway. My concern with the dictionary is that it doesn't suit the spreadsheet requirement, which has to be a sequence, not a dictionary. You're going to write a serialiser there anyway. Why not just provide this serialised as an array of arrays?

However, this here decision is a view-level decision, not a model-level decision. Make these two decisions independently.

boxoffice/models/line_item.py
shreyas-satish: > the starting point isn't necessarily Jan 1

Sure, in such a case, would sales on Jan 1 be counted against the previous year?

jace: Yes.

boxoffice/models/line_item.py
```diff
@@ -3,10 +3,12 @@
 import itertools
 from decimal import Decimal
 import datetime
-from collections import namedtuple
+from collections import namedtuple, OrderedDict
 from sqlalchemy.sql import select, func
 from sqlalchemy.ext.orderinglist import ordering_list
+from isoweek import Week
 from boxoffice.models import db, JsonDict, BaseMixin, Order, Item, DiscountPolicy, DISCOUNT_TYPE, DiscountCoupon, OrderSession
+from boxoffice.views.utils import reset_time_utc, reset_time_local
```
jace: You can't import from views in models. Violates hierarchy.

boxoffice/models/line_item.py
```diff
@@ -235,36 +237,61 @@ def counts_per_date_per_item(item_collection, user_tz):
...
-    Calculates the percentage difference in sales between today and yesterday
+    Calculates sales per week for items in the given set of item_collection_ids in a given year,
+    in the user's timezone.
     """
-    today = datetime.datetime.now().strftime("%Y-%m-%d")
-    yesterday = (datetime.datetime.now() - datetime.timedelta(days=1)).strftime("%Y-%m-%d")
-    sales = sales_by_date([today, yesterday], user_tz, item_ids)
-    if not sales or not sales[yesterday]:
+    ordered_week_sales = OrderedDict()
+    local_start_at = reset_time_local(Week.monday(Week(year, 1)), user_tz)
```
jace: Does `Week.monday` give you a date or a datetime?

jace: If it's a date, what's the timezone calculation for?

boxoffice/models/line_item.py
```diff
@@ -235,36 +237,61 @@ def counts_per_date_per_item(item_collection, user_tz):
...
+    utc_start_at = reset_time_utc(local_start_at, datetime.datetime.min.time())
+    for year_week in Week.weeks_of_year(year):
+        ordered_week_sales[year_week[1]] = 0
+    number_of_weeks = len(ordered_week_sales)
+    local_end_at = reset_time_local(Week.sunday(Week(year, number_of_weeks)), user_tz, time_stamp=datetime.datetime.max.time())
```
jace: I don't think you should be looking for a max point. You want to query for less than next min point.

boxoffice/models/line_item.py
shreyas-satish: `Week.monday` returns a `date`. This `date` is converted to a `datetime` object in the user's timezone, which is then casted to `utc` (in the next line) for it to be used in the query.

boxoffice/models/line_item.py
```diff
@@ -235,36 +236,58 @@ def counts_per_date_per_item(item_collection, user_tz):
...
         return None
 
-    date_sales = {}
+    start_at = midnight_to_utc(sales_datetime, timezone=timezone)
+    end_at = midnight_to_utc(sales_datetime + datetime.timedelta(days=1), timezone=timezone)
+    sales_on_date = db.session.query('sum').from_statement('''SELECT SUM(final_amount) FROM line_item
+        WHERE status=:status AND ordered_at >= :start_at AND ordered_at < :end_at
+        AND line_item.item_id IN :item_ids
+        ''').params(status=LINE_ITEM_STATUS.CONFIRMED, start_at=start_at, end_at=end_at, item_ids=tuple(item_ids)).first()
+    return sales_on_date[0] if sales_on_date[0] else Decimal(0)
```
jace: Use `.scalar()` instead of `.first()` to avoid the `[0]` lookup. http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.scalar

boxoffice/models/line_item.py
```diff
@@ -235,36 +236,58 @@ def counts_per_date_per_item(item_collection, user_tz):
...
     """
-    today = datetime.datetime.now().strftime("%Y-%m-%d")
-    yesterday = (datetime.datetime.now() - datetime.timedelta(days=1)).strftime("%Y-%m-%d")
-    sales = sales_by_date([today, yesterday], user_tz, item_ids)
-    if not sales or not sales[yesterday]:
+    ordered_week_sales = OrderedDict()
+    for year_week in Week.weeks_of_year(year):
+        ordered_week_sales[year_week[1]] = 0
+    start_at = isoweek_datetime(year, 1)
+    end_at = isoweek_datetime(year+1, 1)
```
jace: Spaces around the `+`. Pyflakes should have caught this.

boxoffice/models/line_item.py
```diff
@@ -235,36 +236,58 @@ def counts_per_date_per_item(item_collection, user_tz):
...
+    Calculates sales per week for items in the given set of item_collection_ids in a given year,
+    in the user's timezone.
     """
-    today = datetime.datetime.now().strftime("%Y-%m-%d")
-    yesterday = (datetime.datetime.now() - datetime.timedelta(days=1)).strftime("%Y-%m-%d")
-    sales = sales_by_date([today, yesterday], user_tz, item_ids)
-    if not sales or not sales[yesterday]:
+    ordered_week_sales = OrderedDict()
+    for year_week in Week.weeks_of_year(year):
+        ordered_week_sales[year_week[1]] = 0
```
jace: Use `year_week.week` for clarity.

boxoffice/models/line_item.py
```diff
@@ -235,36 +236,58 @@ def counts_per_date_per_item(item_collection, user_tz):
...
+        AS sales_week, SUM(final_amount)
+        FROM line_item INNER JOIN item on line_item.item_id = item.id
+        WHERE status=:status AND item_collection_id IN :item_collection_ids
+        AND ordered_at >= :start_at AND ordered_at < :end_at
+        GROUP BY sales_week ORDER BY sales_week;
+        ''')).params(timezone=user_tz, status=LINE_ITEM_STATUS.CONFIRMED,
+        start_at=start_at, end_at=end_at, item_collection_ids=tuple(item_collection_ids)).all()
+
+    for week_sale in week_sales:
+        ordered_week_sales[int(week_sale[0])] = week_sale[1]
```
jace: Use named attributes here too. `week_sale.sales_week` and `week_sale.sum`.

boxoffice/models/line_item.py
```diff
@@ -235,36 +236,58 @@ def counts_per_date_per_item(item_collection, user_tz):
...
-    sales = sales_by_date([today, yesterday], user_tz, item_ids)
-    if not sales or not sales[yesterday]:
+    ordered_week_sales = OrderedDict()
+    for year_week in Week.weeks_of_year(year):
+        ordered_week_sales[year_week[1]] = 0
+    start_at = isoweek_datetime(year, 1)
+    end_at = isoweek_datetime(year+1, 1)
+
+    week_sales = db.session.query('sales_week', 'sum').from_statement(db.text('''SELECT EXTRACT(WEEK FROM ordered_at)
+        AS sales_week, SUM(final_amount)
```
jace: Missing a second `AS`.

boxoffice/models/line_item.py
jace: Is the `int()` call necessary? I'd have assumed the data would be integers already.

boxoffice/views/admin_item_collection.py
```diff
@@ -50,8 +51,11 @@ def jsonify_item_collection(item_collection_dict):
 def admin_item_collection(item_collection):
     item_ids = [str(item.id) for item in item_collection.items]
     date_item_counts = counts_per_date_per_item(item_collection, g.user.timezone)
-    date_sales = sales_by_date(date_item_counts.keys(), g.user.timezone, item_ids)
-    today_sales = date_sales.get(datetime.datetime.now().strftime("%Y-%m-%d"), Decimal(0)) if date_sales else Decimal(0)
+    date_sales = {}
+    for date_stamp in date_item_counts.keys():
+        sales_datetime = datetime.datetime.strptime(date_stamp, '%Y-%m-%d')
```
jace: Why does this dictionary have string keys instead of date keys?

jace: The function providing this data is defined in the models. There's no reason for it to be formatting as string only to have it parsed back as date in the view.

boxoffice/views/utils.py
```diff
@@ -24,6 +23,12 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def check_api_access(access_token):
+    """Aborts if a request does not have the correct access token"""
+    if not request.args.get('access_token') or request.args.get('access_token') != access_token:
```
jace: This choice of parameter name adds overhead in Flask-Lastuser, which also attempts to check the access token. https://github.com/hasgeek/flask-lastuser/blob/master/flask_lastuser/__init__.py#L130

boxoffice/models/line_item.py
shreyas-satish: It's necessary since PostgreSQL returns the week number in `float`.

boxoffice/models/line_item.py
```diff
@@ -222,49 +223,72 @@ def counts_per_date_per_item(item_collection, user_tz):
...
         return None
 
-    date_sales = {}
+    start_at = midnight_to_utc(sales_datetime, timezone=timezone)
+    end_at = midnight_to_utc(sales_datetime + datetime.timedelta(days=1), timezone=timezone)
+    sales_on_date = db.session.query('sum').from_statement('''SELECT SUM(final_amount) FROM line_item
+        WHERE status=:status AND ordered_at >= :start_at AND ordered_at < :end_at
+        AND line_item.item_id IN :item_ids
+        ''').params(status=LINE_ITEM_STATUS.CONFIRMED, start_at=start_at, end_at=end_at, item_ids=tuple(item_ids)).scalar()
+    return sales_on_date if sales_on_date else Decimal(0)
```
jace: `return sales_on_date or Decimal(0)` should work to replace `None` with `0`.

boxoffice/models/line_item.py
```diff
@@ -222,49 +223,72 @@ def counts_per_date_per_item(item_collection, user_tz):
...
+    for week_sale in week_sales:
+        ordered_week_sales[int(week_sale.sales_week)] = week_sale.sum
+
+    return ordered_week_sales
+
+
+def sales_delta(user_tz, item_ids):
+    """Calculates the percentage difference in sales between today and yesterday"""
+    today = midnight_to_utc(datetime.datetime.utcnow(), timezone=user_tz)
+    yesterday = midnight_to_utc(datetime.datetime.utcnow() - datetime.timedelta(days=1), timezone=user_tz)
```
jace: Rather than call `utcnow()` twice and get slightly different values, call it once and assign to a variable, and use the variable both times. Convention is to name this variable `now`. Doing this saves you from a suble timing attack when you're milliseconds from midnight and the two calls end up on different days.

boxoffice/models/line_item.py
shreyas-satish: Sure, but I remember we discussed using a one-liner `if/else` instead of an `or`, in an earlier PR. Would you prefer using `or` in this case?

boxoffice/models/line_item.py
shreyas-satish: Ah, that's interesting. Fixed.

boxoffice/models/line_item.py
jace: Do you recall the context?

shreyas-satish: I tried this finding this exact conversation, but I think it was either for a variable assignment or a return expression.

PR #133, opened 2017-02-13
Admin page for discount policy

vidya-ram:
This is PR is to introduce admin page for discount policy.
This will allow:
1. Viewing all discount policy in an organization
2. Adding a new discount policy
3. Editing a  existing discount policy
4. Generating discount coupons.

Dependency:
https://github.com/hasgeek/baseframe/pull/132

small


boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -0,0 +1,512 @@
...
+              templateSelection: function(item) {
+                return item.title;
+              }
+            });
+          }
+        },
+        showNewPolicyForm: function(event) {
+          discountPolicyComponent.set('showAddPolicyForm', true);
+          discountPolicyComponent.set('newDiscountPolicy.is_price_based', 1);
+          discountPolicyComponent.set('newDiscountPolicy.discount_type', 1);
```
shreyas-satish: Perhaps better to store these values (eg: 1) as constants, document them and use those constants, instead of hard-coding the values here.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,217 @@
...
+@app.route('/admin/discount_policy/<discount_policy_id>/edit', methods=['OPTIONS', 'POST'])
+@lastuser.requires_login
+@load_models(
+    (DiscountPolicy, {'id': 'discount_policy_id'}, 'discount_policy'),
+    permission='org_admin'
+    )
+@xhr_only
+def admin_edit_discount_policy(discount_policy):
+    discount_policy_form = DiscountForm()
+    discount_policy_form.items.query = Item.query.all()
```
shreyas-satish: This might load far too many items.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,217 @@
...
+    if not discount_policy_form.validate_on_submit():
+        return make_response(jsonify(status='error', error='invalid_details', error_description=discount_policy_form.errors), 400)
+    if not discount_policy_form.discount_type.data:
+        discount_policy_form.discount_code_base.data = None
+    discount_policy_form.populate_obj(discount_policy)
+    items = request.form.getlist('items')
+    for item_id in items:
+        item = Item.query.get(item_id)
```
shreyas-satish: Given a set of `item_ids` you can fetch all the items at once, instead of running the query in a loop.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,217 @@
...
+        discount_policy.items.append(item)
+    db.session.add(discount_policy)
+    db.session.commit()
+    if discount_policy_form.is_price_based.data:
+        discount_price = Price.query.filter_by(discount_policy=discount_policy).first()
+        discount_price.start_at = discount_policy_form.start_at.data
+        discount_price.end_at = discount_policy_form.end_at.data
+        discount_price.amount = discount_policy_form.amount.data
+        db.session.add(discount_price)
```
shreyas-satish: This isn't required.

boxoffice/views/admin_discount.py
shreyas-satish: `Item.query.join(ItemCollection).filter(ItemCollection.organization == organization).options(db.load_only('item.id', 'item.name', 'item.title')).all()

boxoffice/views/admin_discount.py
shreyas-satish: `Item.query.filter(Item.id.in_(items)).all()`

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,217 @@
...
+    if not discount_policy_form.validate_on_submit():
+        return make_response(jsonify(status='error', error='invalid_details', error_description=discount_policy_form.errors), 400)
+    if not discount_policy_form.discount_type.data:
+        discount_policy_form.discount_code_base.data = None
+    discount_policy_form.populate_obj(discount_policy)
+    items = request.form.getlist('items')
+    for item_id in items:
+        item = Item.query.get(item_id)
+        discount_policy.items.append(item)
+    db.session.add(discount_policy)
```
shreyas-satish: Not required.

boxoffice/views/admin.py
```diff
@@ -40,6 +41,20 @@ def org(organization):
...
+@lastuser.requires_login
+@load_models(
+    (Organization, {'name': 'org'}, 'organization'),
+    permission='org_admin'
+    )
+@xhr_only
+@requestargs('search')
+def admin_items(organization, search=None):
+    if search:
+        filtered_items = Item.query.filter(Item.title.ilike('%{query}%'.format(query=search))).all()
```
shreyas-satish: This query needs to be scoped by the organization.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,217 @@
...
+@xhr_only
+def admin_discount_coupons(discount_policy):
+    discount_coupons = DiscountCoupon.query.filter(DiscountCoupon.discount_policy == discount_policy).all()
+    coupons_list = []
+    for coupon in discount_coupons:
+        coupons_list.append({'code': coupon.code, 'usage_limit': coupon.usage_limit, 'available': coupon.usage_limit - coupon.used_count})
+    return make_response(jsonify(status='ok', result={'message': 'Discount coupons', 'coupons': coupons_list}), 201)
+
+
+@app.route('/admin/discount_policy')
```
shreyas-satish: `/api/1/admin/discount_policy`

vidya-ram: Fixed

boxoffice/static/js/views/admin_discount_policy.js
vidya-ram: Fixed

boxoffice/views/admin.py
vidya-ram: Fixed.

boxoffice/views/admin_discount.py
vidya-ram: Fixed.

boxoffice/views/admin_discount.py
vidya-ram: Fixed.

boxoffice/views/admin_discount.py
vidya-ram: Fixed.

boxoffice/views/admin_discount.py
vidya-ram: Fixed.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,60 @@
...
+        timezone=lambda: g.user.timezone,
+        validators=[forms.validators.OptionalIfNot('is_price_based', (u"Please specify start date for the price"))])
+    end_at = forms.DateTimeField(__("Price end date"), format='%d %b %Y %H:%M:%S', timezone=lambda: g.user.timezone,
+        validators=[forms.validators.OptionalIfNot('is_price_based', (u"Please specify end date for the price"))])
+    items = QuerySelectMultipleField(__("Items"), get_label='title', query_factory=lambda: [],
+        validators=[forms.validators.DataRequired((u"Please select a item to which discount is to be applied"))])
+
+    def validate_discount_code_base(self, field):
+        discount_policy = DiscountPolicy.query.filter_by(discount_code_base=field.data).first()
+        if discount_policy and discount_policy != DiscountPolicy.query.filter_by(title=self.title.data).first():
```
shreyas-satish: Needs scoping by organization.

boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -0,0 +1,491 @@
...
+            }
+          ],
+          getDiscountedItems: function (dpItems) {
+            let discountedItems = dpItems.map(function(dpItem) {
+              return dpItem.id;
+            });
+            return discountedItems.join(',');
+          },
+          getCsrfToken: function () {
+            return document.head.querySelector("[name=csrf-token]").content;
```
shreyas-satish: This can go into `utils.js`.

boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -0,0 +1,491 @@
...
+          }
+          NProgress.start();
+          DiscountPolicyModel.fetch({
+            url: url
+          }).done((remoteData) => {
+            discountPolicyComponent.set('discountPolicies', remoteData.discount_policies);
+            discountPolicyComponent.set('paginated', remoteData.paginated);
+            discountPolicyComponent.set('totalPages', remoteData.total_pages);
+            discountPolicyComponent.set('currentPage', remoteData.current_page);
+            discountPolicyComponent.set('pages', discountPolicyComponent.get_pages(remoteData.total_pages));
```
shreyas-satish: Using an object to set the attributes is probably cleaner.

boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -0,0 +1,491 @@
...
+            NProgress.done();
+          });
+          window.history.replaceState({reloadOnPop: true}, '', window.location.href);
+          window.history.pushState({reloadOnPop: true}, '', url);
```
shreyas-satish: These two lines need to be abstracted.

boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -0,0 +1,491 @@
...
+        },
+        get_pages: function (totalPages) {
+          var pages = [];
+          for (var i=1; i<= totalPages; i++) {
+            pages.push(i);
+          }
+          return pages;
```
shreyas-satish: Use `_.range`

boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -0,0 +1,491 @@
...
+                  };
+                },
+                processResults: function (data) {
+                  return {
+                    results: data.items
+                  };
+                },
+                cache: true
+              },
+              escapeMarkup: function (markup) { return markup; },
```
shreyas-satish: no-op.

boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -0,0 +1,491 @@
...
+            });
+          }
+        },
+        showNewPolicyForm: function (event) {
+          discountPolicyComponent.set('showAddPolicyForm', DEFAULT.showForm);
+          discountPolicyComponent.set('newDiscountPolicy.is_price_based', DEFAULT.priceBasedDiscount);
+          discountPolicyComponent.set('newDiscountPolicy.discount_type', DEFAULT.couponBasedDiscount);
+          discountPolicyComponent.addFormFields(discountPolicyComponent.get('newDiscountPolicy.is_price_based'));
+        },
+        policyChange: function (event) {
```
shreyas-satish: Use the `on` prefix for functions that are event handlers.

boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -0,0 +1,491 @@
...
+
+              DiscountPolicyModel.post({
+                url: DiscountPolicyModel.urlFor('new', {org_name})['path'],
+                data: DiscountPolicyModel.getFormParameters(formSelector)
+              }).done((remoteData) => {
+                discountPolicyComponent.set('discountPolicies', [remoteData.result.discount_policy]);
+                discountPolicyComponent.set('searchText', discountPolicyComponent.get('newDiscountPolicy.title'));
+                discountPolicyComponent.set('newDiscountPolicy.creatingPolicy', DEFAULT.hideLoader);
+                discountPolicyComponent.hideNewPolicyForm();
+                discountPolicyComponent.set('newDiscountPolicy', DEFAULT.empty);
```
shreyas-satish: Use an object `set`.

vidya-ram: Fixed

boxoffice/static/js/views/admin_discount_policy.js
vidya-ram: Fixed

boxoffice/static/js/views/admin_discount_policy.js
vidya-ram: We don't to escape markup since search results are displayed in html

boxoffice/static/js/views/admin_discount_policy.js
vidya-ram: Fixed

boxoffice/static/js/views/admin_discount_policy.js
vidya-ram: Fixed. Moved it utils.js

boxoffice/static/js/views/admin_discount_policy.js
vidya-ram: Done

boxoffice/static/js/views/admin_discount_policy.js
vidya-ram: Done

boxoffice/forms/discount.py
vidya-ram: Is it required? discount_code_base has a unique constraint in the discount policy table not just in the organisation.

shreyas-satish: @vidya-ram The query in Line 46 needs scoping, not Line 45.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,63 @@
...
+        validators=[forms.validators.OptionalIfNot('is_price_based', (u"Please specify end date for the price"))])
+    items = QuerySelectMultipleField(__("Items"), get_label='title', query_factory=lambda: [],
+        validators=[forms.validators.DataRequired((u"Please select a item to which discount is to be applied"))])
+    # For validate_discount_code_base
+    dp_name = forms.StringField(__("Name"),
+        validators=[forms.validators.Optional()], default=None)
+
+    def validate_discount_code_base(self, field):
+        discount_policy = DiscountPolicy.query.filter_by(discount_code_base=field.data).first()
+        if discount_policy and discount_policy.name != self.dp_name.data:
```
shreyas-satish: `name` isn't unique in `DiscountPolicy`. `organization_id` + `name` is.

boxoffice/views/admin.py
```diff
@@ -40,22 +41,35 @@ def org(organization):
...
+@lastuser.requires_login
+@load_models(
+    (Organization, {'name': 'org'}, 'organization'),
+    permission='org_admin'
+    )
+@xhr_only
+@requestargs('search')
+def admin_items(organization, search=None):
+    if search:
+        filtered_items = Item.query.join(ItemCollection).filter(Item.title.ilike('%{query}%'.format(query=search))).all()
```
shreyas-satish: This needs an additional filter to filter it by item collection i.e `.filter(ItemCollection.organization == organization)`.

boxoffice/views/admin.py
```diff
@@ -40,22 +41,35 @@ def org(organization):
...
+        return jsonify(items=[{'id': str(item.id), 'title': item.title} for item in filtered_items])
```
shreyas-satish: Can we use `api_success` and `api_error` functions here as well? `api_error` if the search parameter is `None`.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,212 @@
...
+        'discount_type': "Automatic" if policy.is_automatic else "Coupon based",
+        'item_quantity_min': policy.item_quantity_min,
+        'percentage': policy.percentage,
+        'is_price_based': policy.is_price_based,
+        'discount_code_base': policy.discount_code_base,
+        'bulk_coupon_usage_limit': policy.bulk_coupon_usage_limit,
+        'discount': policy.percentage if not policy.is_price_based else '',
+        'price_details': jsonify_price(Price.query.filter(Price.discount_policy == policy).first()) if policy.is_price_based else '',
+        'currency': CURRENCY_SYMBOL['INR'],
+        'dp_items': [{'id': str(item.id), 'title': item.title} for item in policy.items]
```
shreyas-satish: Probably won't make a significant difference, but generally, it's better to only load the attributes that you are going to use, which in this case, are `id` and `title`.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,212 @@
...
+def admin_add_discount_policy(organization):
+    discount_policy_form = DiscountPolicyForm()
+    discount_policy_form.items.query = Item.query.join(ItemCollection).filter(ItemCollection.organization == organization).options(db.load_only('id')).all()
+    if not discount_policy_form.validate_on_submit():
+        return api_error(message=discount_policy_form.errors, status_code=400)
+    discount_policy = DiscountPolicy(title=discount_policy_form.title.data, organization=organization)
+    discount_policy_form.populate_obj(discount_policy)
+    items = request.form.getlist('items')
+    for item_id in items:
+        item = Item.query.get(item_id)
```
shreyas-satish: I suggest renaming `items` to `item_ids` since it's a list of item ids. This can also be done a little more efficiently. You can fetch all the items at once: `items = Item.query.filter(Item.id.in_(item_ids)).`

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,212 @@
...
+        discount_policy.items.append(item)
+    if discount_policy_form.discount_type.data:
+        discount_policy.secret = buid()
```
shreyas-satish: You can use `discount_policy.set_secret()`.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,212 @@
...
+    if not discount_policy_form.discount_type.data:
+        discount_policy_form.discount_code_base.data = None
+    discount_policy_form.populate_obj(discount_policy)
+    item_ids = request.form.getlist('items')
+    items = Item.query.filter(Item.id.in_(item_ids)).all()
+    discount_policy.items.extend(items)
+    db.session.commit()
+    if discount_policy_form.is_price_based.data:
+        discount_price = Price.query.filter_by(discount_policy=discount_policy).first()
+        discount_price.start_at = discount_policy_form.start_at.data
```
shreyas-satish: What if there's no `discount_price`?

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,212 @@
...
+
+@app.route('/api/1/admin/<org>/discount_policy/<discount_policy_id>/coupons')
+@lastuser.requires_login
+@load_models(
+    (DiscountPolicy, {'id': 'discount_policy_id'}, 'discount_policy'),
+    permission='org_admin'
+    )
+@xhr_only
+def admin_discount_coupons(discount_policy):
+    discount_coupons = DiscountCoupon.query.filter(DiscountCoupon.discount_policy == discount_policy).all()
```
shreyas-satish: This can just be `discount_policy.discount_coupons`.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,212 @@
...
+    for coupon in discount_coupons:
+        coupons_list.append({'code': coupon.code, 'usage_limit': coupon.usage_limit, 'available': coupon.usage_limit - coupon.used_count})
+    return api_success(result={'coupons': coupons_list}, doc="List of discount coupons", status_code=200)
+
+
+@app.route('/api/1/admin/discount_policy')
+@lastuser.requires_login
+@xhr_only
+@requestargs('discount_code_base')
+def admin_discount_policy(discount_code_base):
```
shreyas-satish: What is this method for? Think it needs a better name. Also needs a permission check.

vidya-ram: Removed now.

boxoffice/views/admin_discount.py
vidya-ram: Fixed

boxoffice/views/admin_discount.py
vidya-ram: Fixed

boxoffice/views/admin_discount.py
vidya-ram: Fixed

boxoffice/views/admin_discount.py
vidya-ram: Fixed

boxoffice/views/admin.py
vidya-ram: Changed

boxoffice/views/admin.py
vidya-ram: Fixed

boxoffice/forms/discount.py
vidya-ram: Fixed

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,197 @@
...
+    items = Item.query.filter(Item.id.in_(item_ids))
+    discount_policy.items.extend(items)
+    if discount_policy_form.discount_type.data:
+        discount_policy.set_secret()
+    else:
+        discount_policy_form.discount_code_base.data = None
+    db.session.add(discount_policy)
+    db.session.commit()
+    if discount_policy_form.is_price_based.data:
+        discount_price = Price(discount_policy=discount_policy, title=discount_policy_form.price_title.data, start_at=discount_policy_form.start_at.data, end_at=discount_policy_form.end_at.data, amount=discount_policy_form.amount.data, item=items[0])
```
shreyas-satish: Maybe better to use the form object's `populate_obj` method to fill in values?

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,197 @@
...
+    item_ids = request.form.getlist('items')
+    items = Item.query.filter(Item.id.in_(item_ids))
+    discount_policy.items.extend(items)
+    if discount_policy_form.discount_type.data:
+        discount_policy.set_secret()
+    else:
+        discount_policy_form.discount_code_base.data = None
+    db.session.add(discount_policy)
+    db.session.commit()
+    if discount_policy_form.is_price_based.data:
```
shreyas-satish: Just `discount_policy.is_price_based` should be sufficient?

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,197 @@
...
+        return api_error(message=discount_policy_form.errors, status_code=400)
+    discount_policy = DiscountPolicy(title=discount_policy_form.title.data, organization=organization)
+    discount_policy_form.populate_obj(discount_policy)
+    item_ids = request.form.getlist('items')
+    items = Item.query.filter(Item.id.in_(item_ids))
+    discount_policy.items.extend(items)
+    if discount_policy_form.discount_type.data:
+        discount_policy.set_secret()
+    else:
+        discount_policy_form.discount_code_base.data = None
```
shreyas-satish: Why is this being set here? Since `populate_obj` has already been called in #L104.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,197 @@
...
+        discount_policies = discount_policies.limit(results_per_page).offset(offset).all()
+
+        return dict(
+            org=organization, title=organization.title,
+            discount_policies=discount_policies,
+            total_pages=total_pages,
+            paginated=(total_policies > results_per_page),
+            current_page=page)
+    else:
+        return dict(org=organization, title=organization.title)
```
shreyas-satish: Since `jsonify_discount_policies` only seems to need `organization.name` and `organization.title`, maybe just return those two attributes only, instead of returning the entire object?

boxoffice/views/admin.py
```diff
@@ -40,22 +41,37 @@ def org(organization):
...
+@lastuser.requires_login
+@load_models(
+    (Organization, {'name': 'org'}, 'organization'),
+    permission='org_admin'
+    )
+@xhr_only
+@requestargs('search')
+def filter_items(organization, search=None):
+    if search:
+        filtered_items = Item.query.join(ItemCollection).filter(ItemCollection.organization == organization).filter(Item.title.ilike('%{query}%'.format(query=search))).all()
```
shreyas-satish: Perhaps break this line up into two lines? It becomes hard to read if it's too long.

boxoffice/views/admin_discount.py
vidya-ram: Fixed

boxoffice/views/admin_discount.py
vidya-ram: discount_policy_form.is_price_based.data contains the value.

boxoffice/views/admin_discount.py
vidya-ram: Moved this before populate_obj

boxoffice/views/admin_discount.py
vidya-ram: Fixed

boxoffice/views/admin.py
vidya-ram: Done

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,195 @@
...
+    if not discount_policy_form.validate_on_submit():
+        return api_error(message=discount_policy_form.errors, status_code=400)
+    discount_policy = DiscountPolicy(title=discount_policy_form.title.data, organization=organization)
+    item_ids = request.form.getlist('items')
+    items = Item.query.filter(Item.id.in_(item_ids))
+    discount_policy.items.extend(items)
+    if discount_policy_form.discount_type.data:
+        discount_policy.set_secret()
+    else:
+        discount_policy_form.discount_code_base.data = None
```
shreyas-satish: Does this _have_ to be set to `None`?

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,195 @@
...
+    discount_policy_form.populate_obj(discount_policy)
+    db.session.add(discount_policy)
+    db.session.commit()
+    if discount_policy_form.is_price_based.data:
```
shreyas-satish: Shouldn't this be `if discount_policy.is_price_based`?

vidya-ram: Udpated

boxoffice/views/admin_discount.py
vidya-ram: wtf form sets the optional field to empty string, so it was being set here to None.
Added a filter to the form validation to set it None incase it is empty.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+    discount_policy.make_name()
+    db.session.add(discount_policy)
+    db.session.commit()
+
+    if discount_policy.is_price_based:
+        discount_price = Price(discount_policy=discount_policy)
+        discount_price_form.populate_obj(discount_price)
+        discount_price.make_name()
+        db.session.add(discount_price)
+        discount_policy.items.append(discount_price.item)
```
jace: We have a database normalisation problem here. `DiscountPolicy` and `Price` both link to `Item` and also to each other, which means they could potentially link to unrelated `Items`. One of the models should have a pre-commit validator to ensure this does not happen.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+        db.session.commit()
+    return api_success(result={'discount_policy': jsonify_discount_policy(discount_policy)},
+        doc="New discount policy created.", status_code=200)
```
jace: Return `201` since this is creating a new resource.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+    (Organization, {'name': 'org'}, 'organization'),
+    (DiscountPolicy, {'id': 'discount_policy_id'}, 'discount_policy'),
+    permission='org_admin'
+    )
+@xhr_only
+def admin_edit_discount_policy(organization, discount_policy):
+    discount_policy_form = DiscountPolicyForm(request.form)
+    all_items = Item.query.join(ItemCollection).filter(
+        ItemCollection.organization == organization).options(db.load_only('id')).all()
+    discount_policy_form.items.query = all_items
```
jace: If the query is moved into the form, we can avoid reproducing it here.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+
+    if not discount_policy_form.validate_on_submit():
+        return api_error(message=discount_policy_form.errors, status_code=400)
+
+    if not discount_policy.is_price_based:
+        discount_policy.items.extend(discount_policy_form.items.data)
```
jace: Shouldn't this replace rather than extend? How do you remove the items it's associated with?

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+        discount_policy_form.populate_obj(discount_policy)
+    else:
+        discount_price_form = DiscountPriceForm(request.form)
+        discount_price_form.item.query = all_items
```
jace: As above.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+        if not discount_price_form.validate_on_submit():
+            return api_error(message=discount_price_form.errors, status_code=400)
+        discount_price = Price.query.filter_by(discount_policy=discount_policy).one_or_none()
+        if discount_price:
+            discount_price_form.populate_obj(discount_price)
+            discount_policy_form.populate_obj(discount_policy)
```
jace: We're not updating `discount_policy` if there's no price? There's no error shown to user either.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+    return api_success(result={'discount_policy': jsonify_discount_policy(discount_policy)}, doc="Discount policy updated.", status_code=200)
+
+
+@app.route('/admin/discount_policy/<discount_policy_id>/coupons/new', methods=['OPTIONS', 'POST'])
+@lastuser.requires_login
+@load_models(
+    (DiscountPolicy, {'id': 'discount_policy_id'}, 'discount_policy'),
+    permission='org_admin'
+    )
+@xhr_only
```
jace: Move up.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+def admin_create_coupon(discount_policy):
+    coupon_form = DiscountCouponForm()
+    coupons = []
+    if not coupon_form.validate_on_submit():
+        return api_error(message=coupon_form.errors, status_code=400)
```
jace: This is incorrect use of `message`. The message should always be a human readable message, with a separate `errors` parameter enumerating the fields and error messages.

jace: @shreyas-satish you may want to fix this in the `api_error` function.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+                usage_limit=coupon_form.usage_limit.data,
+                code=coupon_form.coupon_code.data)
+        else:
+            # Randomly generated discount code
+            coupon = DiscountCoupon(discount_policy=discount_policy,
+                usage_limit=coupon_form.usage_limit.data)
+        db.session.add(coupon)
+        db.session.commit()
+        coupons.append(coupon.code)
+    return api_success(result={'coupons': coupons}, doc="Discount coupon created.", status_code=200)
```
jace: Status `201`

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+@app.route('/admin/discount_policy/<discount_policy_id>/coupons')
+@lastuser.requires_login
+@load_models(
+    (DiscountPolicy, {'id': 'discount_policy_id'}, 'discount_policy'),
+    permission='org_admin'
+    )
+@xhr_only
```
jace: Move up

boxoffice/views/admin_order.py
```diff
@@ -72,12 +72,13 @@ def admin_orders(item_collection):
...
 
 
-@app.route('/admin/order/<order_id>')
+@app.route('/admin/ic/<ic_id>/order/<order_id>')
 @lastuser.requires_login
 @load_models(
     (Order, {'id': 'order_id'}, 'order'),
     permission='org_admin'
     )
+@xhr_only
```
jace: Move up

boxoffice/views/admin_order.py
```diff
@@ -72,12 +72,13 @@ def admin_orders(item_collection):
     return dict(title=item_collection.organization.title, item_collection=item_collection, orders=item_collection.orders)
 
 
-@app.route('/admin/order/<order_id>')
+@app.route('/admin/ic/<ic_id>/order/<order_id>')
```
jace: Why take an `ic_id` parameter if you're not using it?

boxoffice/views/admin_report.py
```diff
@@ -8,24 +8,25 @@
 from boxoffice.views.utils import csv_response
 from babel.dates import format_datetime
 
+
 def jsonify_report(data_dict):
-    return jsonify(org_name=data_dict['item_collection'].organization.name, item_collection_title=data_dict['item_collection'].title)
+    return jsonify(org_name=data_dict['item_collection'].organization.name, title=data_dict['item_collection'].title)
```
jace: If `title` isn't unique to the organization, this doesn't seem the appropriate set of values to return.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+        validators=[forms.validators.OptionalIf('is_price_based', (u"Please select a item to which discount is to be applied"))])
+    # For validate_discount_code_base
+    discount_policy_id = forms.StringField(__("Id"),
+        validators=[forms.validators.Optional()], default=None)
+
+    def validate_discount_code_base(self, field):
+        if self.discount_policy_id.data:
+            if DiscountPolicy.query.filter(DiscountPolicy.id != self.discount_policy_id.data, DiscountPolicy.discount_code_base == field.data).notempty():
+                raise forms.ValidationError((u"Please specify a different discount code base"))
+        elif DiscountPolicy.query.filter_by(discount_code_base=field.data).notempty():
```
shreyas-satish: This is because discount policies were designed to work for items across item collections. Although, in hindsight, `discount_code_base` should perhaps have been designed to be unique per organization.

boxoffice/__init__.py
```diff
@@ -41,7 +41,7 @@ def init_for(env):
     lastuser.init_app(app)
     lastuser.init_usermanager(UserManager(db, User))
     app.config['tz'] = timezone(app.config['TIMEZONE'])
-    baseframe.init_app(app, requires=['boxoffice'], ext_requires=['baseframe-bs3', 'fontawesome>=4.0.0', 'ractive', 'ractive-transitions-fly', 'validate', 'nprogress', 'baseframe-footable'])
+    baseframe.init_app(app, requires=['boxoffice'], ext_requires=['baseframe-bs3-select2', 'fontawesome>=4.0.0', 'ractive', 'ractive-transitions-fly', 'validate', 'nprogress', 'baseframe-footable'])
```
vidya-ram: Fixed

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+import baseframe.forms as forms
+from baseframe.forms.sqlalchemy import QuerySelectMultipleField, QuerySelectField
+from coaster.utils import getbool
+from ..models import DISCOUNT_TYPE, DiscountPolicy, CURRENCY
+
+__all__ = ['DiscountPolicyForm', 'DiscountPriceForm', 'DiscountCouponForm']
+
+
+class DiscountPolicyForm(forms.Form):
+    title = forms.StringField(__("Discount Title"),
```
vidya-ram: Fixed

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+        validators=[forms.validators.DataRequired((u"Please specify a discount title")),
```
vidya-ram: Updated

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+        forms.validators.Length(max=250), forms.validators.StripWhitespace()])
+    is_price_based = forms.RadioField(__("Price based discount"),
+        choices=[(True, 'Special price discount'),
+        (False, 'Percentage based discount')], coerce=getbool, default=True)
```
vidya-ram: Fixed

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+    discount_type = forms.RadioField(__("Discount Type"),
+        choices=DISCOUNT_TYPE.items(), coerce=int, default=1)
```
vidya-ram: Fixed

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+        validators=[forms.validators.DataRequired((u"Please specify a title for the discount price")),
+        forms.validators.Length(max=250), forms.validators.StripWhitespace()])
+    is_price_based = forms.RadioField(__("Price based discount"),
+        choices=[(True, 'Special price discount'),
+        (False, 'Percentage based discount')], coerce=getbool, default=True)
+    amount = forms.IntegerField(__("Amount"),
+        validators=[forms.validators.OptionalIfNot('is_price_based', (u"Please specify an amount"))])
+    currency = forms.RadioField(__("Currency"),
+        validators=[forms.validators.OptionalIfNot('is_price_based', (u"Please select the currency"))],
+        choices=CURRENCY.items(), default='INR')
```
vidya-ram: Fixed

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+            raise forms.ValidationError((u"Please specify the end date for the price that is greater than start date"))
+
+
+class DiscountCouponForm(forms.Form):
+    count = forms.IntegerField(__("Number of coupons to be generated"), default=1)
+    usage_limit = forms.IntegerField(__("Number of times each coupon can be used"), default=1)
+    coupon_code = forms.StringField(__("Code for discount coupon"),
+        validators=[forms.validators.Optional(),
+        forms.validators.Length(max=100),
+        forms.validators.StripWhitespace()], default='')
```
vidya-ram: Fixed

boxoffice/templates/layout.html
```diff
@@ -12,6 +12,7 @@
 {% endblock %}
 
 {% block headerbox %}
+  <link rel="stylesheet" type="text/css" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css">
```
vidya-ram: Fixed

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+    )
+
+
+@app.route('/admin/o/<org>/discount_policy')
+@lastuser.requires_login
+@load_models(
+    (Organization, {'name': 'org'}, 'organization'),
+    permission='org_admin'
+    )
+@render_with({'text/html': 'index.html', 'application/json': jsonify_discount_policies}, json=True)
```
vidya-ram: Fixed

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+        return dict(title=organization.title)
+
+
+@app.route('/admin/o/<org>/discount_policy/new', methods=['OPTIONS', 'POST'])
+@lastuser.requires_login
+@load_models(
+    (Organization, {'name': 'org'}, 'organization'),
+    permission='org_admin'
+    )
+@xhr_only
```
vidya-ram: Fixed

boxoffice/views/admin_discount.py
vidya-ram: Fixed

boxoffice/views/admin_discount.py
vidya-ram: Fixed

boxoffice/views/admin_discount.py
vidya-ram: Fixed

boxoffice/views/admin_discount.py
vidya-ram: Fixed

boxoffice/views/admin_order.py
vidya-ram: Done

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+        forms.validators.Length(max=250), forms.validators.StripWhitespace()])
+    is_price_based = forms.RadioField(__("Price based discount"),
+        choices=[(True, 'Special price discount'),
+        (False, 'Percentage based discount')], coerce=getbool, default=True)
+    amount = forms.IntegerField(__("Amount"),
+        validators=[forms.validators.OptionalIfNot('is_price_based', (u"Please specify an amount"))])
+    currency = forms.RadioField(__("Currency"),
+        validators=[forms.validators.OptionalIfNot('is_price_based', (u"Please select the currency"))],
+        choices=CURRENCY.items(), default='INR')
+    start_at = forms.DateTimeField(__("Price start date"), format='%d %b %Y %H:%M:%S',
```
vidya-ram: Baseframe default format is %Y-%m-%d %H:%M. We would need seconds to for prices.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+        timezone=lambda: g.user.timezone,
```
vidya-ram: Updated

boxoffice/forms/discount.py
jace: We shouldn't need seconds. We changed the code so it doesn't need 23:59:59 anymore.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+@load_models(
+    (Organization, {'name': 'org'}, 'organization'),
+    permission='org_admin'
+    )
+@render_with({'text/html': 'index.html', 'application/json': jsonify_discount_policies}, json=True)
+@requestargs('search', ('page', int))
+def admin_discount_policies(organization, search=None, page=1):
+    results_per_page = 20
+
+    if request.is_xhr:
```
shreyas-satish: Fixed.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+        discount_policies = discount_policies.limit(results_per_page).offset(offset).all()
+
+        return dict(
+            org=organization, title=organization.title,
+            discount_policies=discount_policies,
+            total_pages=total_pages,
+            paginated=(total_policies > results_per_page),
+            current_page=page)
+    else:
+        return dict(title=organization.title)
```
shreyas-satish: The `if` has been removed.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,65 @@
...
+    discount_type = forms.RadioField(__("Discount Type"),
+        choices=DISCOUNT_TYPE.items(), coerce=int, default=DISCOUNT_TYPE.values()[1])
+    percentage = forms.IntegerField(__("Percentage"),
+        validators=[forms.validators.OptionalIf('is_price_based', __("Please specify a discount percentage"))])
+    item_quantity_min = forms.IntegerField(__("Minimum number of tickets"), default=1)
+    discount_code_base = forms.NullTextField(__("Discount Title"),
+        validators=[forms.validators.OptionalIfNot('discount_type', __("Please specify a discount code base")),
+        forms.validators.Length(max=20), forms.validators.StripWhitespace()])
+    bulk_coupon_usage_limit = forms.IntegerField(__("Number of times a bulk coupon can be used"), default=1)
+    items = QuerySelectMultipleField(__("Items"), get_label='title', query_factory=lambda: [],
```
shreyas-satish: Why should this be mandatory only if the discount policy is price based? I think it makes sense to make `items` mandatory regardless.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+        (False, 'Percentage based discount')], coerce=getbool, default=True)
+    discount_type = forms.RadioField(__("Discount Type"),
+        choices=DISCOUNT_TYPE.items(), coerce=int, default=1)
+    percentage = forms.IntegerField(__("Percentage"),
+        validators=[forms.validators.OptionalIf('is_price_based', (u"Please specify a discount percentage"))])
+    item_quantity_min = forms.IntegerField(__("Minimum number of tickets"), default=1)
+    discount_code_base = forms.StringField(__("Discount Title"),
+        validators=[forms.validators.OptionalIfNot('discount_type',
+            (u"Please specify a discount code base")),
+        forms.validators.Length(max=20), forms.validators.StripWhitespace()], filters=[lambda code_base: code_base or None])
```
vidya-ram: Fixed

boxoffice/forms/discount.py
vidya-ram: Fixed

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+    start_at = forms.DateTimeField(__("Price start date"), format='%d %b %Y %H:%M:%S',
+        timezone=lambda: g.user.timezone,
+        validators=[forms.validators.OptionalIfNot('is_price_based', (u"Please specify a start date and time"))])
+    end_at = forms.DateTimeField(__("Price end date"), format='%d %b %Y %H:%M:%S',
+        timezone=lambda: g.user.timezone,
+        validators=[forms.validators.OptionalIfNot('is_price_based', (u"Please specify an end date and time"))])
+    item = QuerySelectField(__("Item"), get_label='title', query_factory=lambda: [],
+        validators=[forms.validators.OptionalIfNot('is_price_based', (u"Please select a item to which the discount is to be applied"))])
+
+    def validate_end_at(self, field):
```
vidya-ram: Fixed

boxoffice/views/admin_discount.py
```diff
@@ -36,7 +36,7 @@ def jsonify_discount_policy(policy):
         'discount': policy.percentage if not policy.is_price_based else '',
         'price_details': jsonify_price(Price.query.filter(Price.discount_policy == policy).first()) if policy.is_price_based else '',
         'currency': CURRENCY_SYMBOL['INR'],
-        'dp_items': [{'id': str(item.id), 'title': item.title} for item in policy.items]
+        'dp_items': [{'id': str(item.id), 'title': "{ic_title}: {title}".format(ic_title=item.item_collection.title, title=item.title)} for item in policy.items]
```
shreyas-satish: This will result in a `n+1` query. Ideally, we'd like to fetch all the data required in as few database hits as possible.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,208 @@
...
+        'discount_type': "Automatic" if policy.is_automatic else "Coupon based",
+        'item_quantity_min': policy.item_quantity_min,
+        'percentage': policy.percentage,
+        'is_price_based': policy.is_price_based,
+        'discount_code_base': policy.discount_code_base,
+        'bulk_coupon_usage_limit': policy.bulk_coupon_usage_limit,
+        'discount': policy.percentage if not policy.is_price_based else '',
+        'price_details': jsonify_price(Price.query.filter(Price.discount_policy == policy).first()) if policy.is_price_based else '',
+        'currency': CURRENCY_SYMBOL['INR'],
+        'dp_items': [{'id': str(item.id), 'title': item.title} for item in policy.items]
```
shreyas-satish: Another case of `n+1`.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+    percentage = forms.IntegerField(__("Percentage"),
+        validators=[forms.validators.OptionalIf('is_price_based', (u"Please specify a discount percentage"))])
+    item_quantity_min = forms.IntegerField(__("Minimum number of tickets"), default=1)
+    discount_code_base = forms.StringField(__("Discount Title"),
+        validators=[forms.validators.OptionalIfNot('discount_type',
+            (u"Please specify a discount code base")),
+        forms.validators.Length(max=20), forms.validators.StripWhitespace()], filters=[lambda code_base: code_base or None])
+    bulk_coupon_usage_limit = forms.IntegerField(__("Number of times a bulk coupon can be used"), default=1)
+    items = QuerySelectMultipleField(__("Items"), get_label='title', query_factory=lambda: [],
+        validators=[forms.validators.OptionalIf('is_price_based', (u"Please select a item to which discount is to be applied"))])
```
shreyas-satish: Fixed.

boxoffice/views/admin_discount.py
shreyas-satish: Added a pre-commit validator.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+@load_models(
+    (Organization, {'name': 'org'}, 'organization'),
+    permission='org_admin'
+    )
+@xhr_only
+def admin_add_discount_policy(organization):
+    discount_policy_form = DiscountPolicyForm(request.form)
+    all_items = Item.query.join(ItemCollection).filter(
+        ItemCollection.organization == organization).options(db.load_only('id')).all()
+    discount_policy_form.items.query = all_items
```
shreyas-satish: Fixed.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+@app.route('/admin/o/<org>/discount_policy')
+@lastuser.requires_login
+@load_models(
+    (Organization, {'name': 'org'}, 'organization'),
+    permission='org_admin'
+    )
+@render_with({'text/html': 'index.html', 'application/json': jsonify_discount_policies}, json=True)
+@requestargs('search', ('page', int))
+def admin_discount_policies(organization, search=None, page=1):
+    results_per_page = 20
```
vidya-ram: Fixed

boxoffice/views/admin_order.py
vidya-ram: Fixed

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+    # For validate_discount_code_base
+    discount_policy_id = forms.StringField(__("Id"),
+        validators=[forms.validators.Optional()], default=None)
```
shreyas-satish: This field has been removed now.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def validate_discount_code_base(self, field):
+        if self.discount_policy_id.data:
```
shreyas-satish: This has been re-written now, to use `edit_id` to get the discount policy id.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,210 @@
...
+
+@app.route('/admin/o/<org>/discount_policy/new', methods=['OPTIONS', 'POST'])
+@lastuser.requires_login
+@load_models(
+    (Organization, {'name': 'org'}, 'organization'),
+    permission='org_admin'
+    )
+@xhr_only
+def admin_add_discount_policy(organization):
+    discount_policy_form = DiscountPolicyForm(request.form)
```
shreyas-satish: Fixed.

boxoffice/views/admin_discount.py
shreyas-satish: This has been fixed because of the re-write.

boxoffice/views/admin_discount.py
shreyas-satish: Fixed.

boxoffice/views/admin_report.py
shreyas-satish: Right, but since organization's `name` is being returned, this shouldn't be a problem?

jace: Where is this used?

boxoffice/static/js/views/router.js
```diff
@@ -3,30 +3,52 @@ import {IndexView} from './index.js';
...
   },
-  org: function(org){
-    OrgView.render({name: org});
+  org: function(org_name) {
+    OrgView.render('view', {org_name});
   },
-  item_collection: function(icId){
-    ItemCollectionView.render({id: icId});
+  item_collection: function(ic_id) {
+    ItemCollectionView.render('dashboard', {ic_id});
```
shreyas-satish: @vidya-ram The `render` calls need to be fixed, both at the caller and the callee, since the first argument isn't being used anymore.

boxoffice/views/admin_report.py
shreyas-satish: The title is used as one part of the report's filename eg: `<title>_tickets.csv`.

boxoffice/static/js/views/router.js
vidya-ram: Fixed

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,224 @@
...
+        'id': policy.id,
+        'name': policy.name,
+        'title': policy.title,
+        'discount_type': "Automatic" if policy.is_automatic else "Coupon based",
+        'item_quantity_min': policy.item_quantity_min,
+        'percentage': policy.percentage,
+        'is_price_based': policy.is_price_based,
+        'discount_code_base': policy.discount_code_base,
+        'bulk_coupon_usage_limit': policy.bulk_coupon_usage_limit,
+        'discount': policy.percentage if not policy.is_price_based else '',
```
shreyas-satish: @vidya-ram Is this attribute required? Since `percentage` and `is_price_based` are being returned above anyway.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,224 @@
...
+@xhr_only
+@load_models(
+    (DiscountPolicy, {'id': 'discount_policy_id'}, 'discount_policy'),
+    permission='org_admin'
+    )
+def admin_new_coupon(discount_policy):
+    coupon_form = DiscountCouponForm(parent=discount_policy)
+    coupons = []
+    if not coupon_form.validate_on_submit():
+        return api_error(message=_(u"The coupon could not be created. Please rectify the indicated issues."),
```
jace: No trailing period.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,224 @@
...
+        for x in range(coupon_form.count.data):
+            # No need to store these coupon codes since they are signed
+            coupons.append(discount_policy.gen_signed_code())
+    else:
+        coupon = DiscountCoupon(discount_policy=discount_policy)
+        coupon_form.populate_obj(coupon)
+        db.session.add(coupon)
+        db.session.commit()
+        coupons.append(coupon.code)
+    return api_success(result={'coupons': coupons}, doc=_(u"Discount coupon created."), status_code=201)
```
jace: No trailing period.

boxoffice/views/admin_item.py
```diff
@@ -0,0 +1,31 @@
...
+@load_models(
+    (Organization, {'name': 'org'}, 'organization'),
+    permission='org_admin'
+    )
+@requestargs('search')
+def items(organization, search=None):
+    if search:
+        filtered_items = db.session.query(Item, ItemCollection).filter(
+            ItemCollection.organization == organization).filter(
+            Item.title.ilike('%{query}%'.format(query=search))).join(Item.item_collection).options(
```
jace: Don't use `LIKE`/`ILIKE` queries with a leading `%`. It forces a full table scan as there is no index for it. This is fine when you have a small number of items, but doesn't scale. Use PostgreSQL's full text indexing instead.

jace: To enable full text indexing, you'll need a migration, plus a custom `before_create` event listener for SQLAlchemy. Look at Hasjob for an example of how to do it.

boxoffice/views/admin_report.py
jace: You may need `coaster.utils.make_name(title)` for the filename, then.

boxoffice/__init__.py
```diff
@@ -7,6 +7,8 @@
 from flask.ext.lastuser import Lastuser
 from flask.ext.lastuser.sqlalchemy import UserManager
 from flask_admin import Admin
+from flask.json import JSONEncoder as BaseEncoder
+from speaklater import _LazyString
```
shreyas-satish: Switched to `is_lazy_string`, which is part of the official API.

boxoffice/__init__.py
```diff
@@ -30,11 +32,19 @@
...
+            return unicode(obj)
+        return BaseEncoder.default(self, obj)
+
+
 # Configure the app
 def init_for(env):
     coaster.app.init_app(app, env)
     db.init_app(app)
     db.app = app
+    app.json_encoder = JSONEncoder
```
shreyas-satish: Issued a PR - https://github.com/hasgeek/baseframe/pull/146

boxoffice/views/admin_report.py
shreyas-satish: Or I can just use `item_collection.name` for the filename?

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,224 @@
...
+        'title': policy.title,
+        'discount_type': "Automatic" if policy.is_automatic else "Coupon based",
+        'item_quantity_min': policy.item_quantity_min,
+        'percentage': policy.percentage,
+        'is_price_based': policy.is_price_based,
+        'discount_code_base': policy.discount_code_base,
+        'bulk_coupon_usage_limit': policy.bulk_coupon_usage_limit,
+        'discount': policy.percentage if not policy.is_price_based else '',
+        'price_details': jsonify_price(Price.query.filter(Price.discount_policy == policy).first()) if policy.is_price_based else '',
+        'currency': CURRENCY_SYMBOL['INR'],
```
shreyas-satish: Fixed.

boxoffice/views/admin_discount.py
shreyas-satish: Fixed.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,100 @@
...
+        super(AutomaticDiscountPolicyForm, self).__init__(*args, **kwargs)
+        self.items.query = Item.query.join(ItemCollection).filter(
+            ItemCollection.organization == self.edit_parent).options(db.load_only('id', 'title'))
+
+
+class CouponBasedDiscountPolicyForm(DiscountPolicyForm):
+    items = QuerySelectMultipleField(__("Items"), get_label='title',
+        validators=[forms.validators.DataRequired(__("Please select at least one item for which the discount is applicable"))])
+    percentage = forms.IntegerField(__("Percentage"),
+        validators=[forms.validators.DataRequired(__("Please specify a discount percentage"))])
```
shreyas-satish: `forms.validators.NumberRange(min=1, max=100)` would suffice?

boxoffice/views/admin_item.py
shreyas-satish: Added a migration that adds a GIN index to `discount_policy.title` which will enable [trigram matching](https://www.postgresql.org/docs/9.6/static/pgtrgm.html).

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,100 @@
...
+        forms.validators.Length(max=250)], filters=[forms.filters.strip()])
+    discount_type = forms.RadioField(__("Discount Type"),
+        choices=DISCOUNT_TYPE.items(), coerce=int, default=DISCOUNT_TYPE.COUPON)
+    is_price_based = forms.RadioField(__("Price based discount"),
+        choices=[(True, __("Special price discount")),
+        (False, __("Percentage based discount"))], coerce=getbool, default=True)
+
+
+def validate_unique_discount_code_base(form, field):
+    if DiscountPolicy.query.filter(DiscountPolicy.id != form.edit_id, DiscountPolicy.discount_code_base == field.data).notempty():
```
shreyas-satish: Done.

boxoffice/static/js/models/util.js
```diff
@@ -19,25 +19,104 @@ export const Util = {
...
+};
 
-export const post = function(config){
+export const post = function (config) {
   return $.ajax({
     url: config.url,
     type: 'POST',
-    dataType: 'json'
+    data: config.data,
+    contentType : config.contentType ? config.contentType : 'application/x-www-form-urlencoded; charset=UTF-8',
```
shreyas-satish: Since this can either be a JSON object or a form object, it's better for config to accept a `json` flag. That way, the caller of `post` does not have to be aware of jQuery's `ajax` syntax, which should be an internal detail of `post`.

boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -0,0 +1,542 @@
...
+        data:  {
+          org: org_name,
+          title: title,
+          discountPolicies: discount_policies,
+          currency_symbol: currency_symbol,
+          paginated: paginated,
+          totalPages: total_pages,
+          currentPage: current_page,
+          size: 20,
+          items: '',
```
shreyas-satish: Looks like `items` is used as an array, further down? If that's the case, it's better to initialized with `[]`.

boxoffice/static/js/views/admin_discount_policy.js
```diff
@@ -0,0 +1,542 @@
...
+              size: size||discountPolicyComponent.get('size')
+            });
+          } else {
+            url = urlFor('index', {
+              scope_ns: 'o',
+              scope_id: org_name,
+              resource: 'discount_policy',
+              root: true,
+              page: page,
+              size: size||discountPolicyComponent.get('size')
```
shreyas-satish: spaces around `||`.

boxoffice/static/js/models/util.js
```diff
@@ -19,25 +19,109 @@ export const Util = {
...
+export const getFormParameters = function (form) {
+  var form_elements = $(form).serializeArray();
+  return $.param(form_elements);
+};
+
+export const getCsrfToken = function () {
+  return document.head.querySelector("[name=csrf-token]").content;
+};
+
+export const updateBrowserHistory = function (currentUrl, newUrl) {
```
shreyas-satish: Does the `currentUrl` _have_ to be passed, can it be inferred?

vidya-ram: Fixed

boxoffice/static/js/views/admin_discount_policy.js
vidya-ram: Fixed

boxoffice/static/js/views/admin_discount_policy.js
vidya-ram: Fixed

boxoffice/static/js/models/util.js
vidya-ram: Fixed

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,224 @@
...
+from ..forms import DiscountPolicyForm, DiscountCouponForm, DiscountPriceForm, CouponBasedDiscountPolicyForm, AutomaticDiscountPolicyForm, PriceBasedDiscountPolicyForm
+from utils import xhr_only, date_time_format, api_error, api_success
+
+
+def jsonify_price(price):
+    if price:
+        return {
+            'amount': price.amount,
+            'start_at': date_time_format(price.start_at),
+            'end_at': date_time_format(price.end_at)
```
vidya-ram: @jace Could we fix this in a separate PR.

jace: We could move the Baseframe dependency out, but use `isoformat()+'Z'` here.

vidya-ram: Fixed

boxoffice/templates/index.html
```diff
@@ -2,6 +2,7 @@
 
 {%- block layoutheaders %}
   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/c3/0.4.11/c3.css">
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-daterangepicker/2.1.25/daterangepicker.min.css">
```
jace: Should this be in Baseframe? We should also get C3 in there now as we're using it regularly.

boxoffice/templates/layout.html
```diff
@@ -2,13 +2,6 @@
 {% from "baseframe/components.html" import hgnav with context %}
 
 {%- block layoutheaders %}
-  <script>
-      // (function(_,e,rr,s){_errs=[s];var c=_.onerror;_.onerror=function(){var a=arguments;_errs.push(a);
-      // c&&c.apply(this,a)};var b=function(){var c=e.createElement(rr),b=e.getElementsByTagName(rr)[0];
-      // c.src="//beacon.errorception.com/"+s+".js";c.async=!0;b.parentNode.insertBefore(c,b)};
-      // _.addEventListener?_.addEventListener("load",b,!1):_.attachEvent("onload",b)})
-      // (window,document,"script","56f13edc2777d31309000c21");
-  </script>
```
jace: This should be a separate PR/commit.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,223 @@
...
+
+    if search:
+        discount_policies = discount_policies.filter(
+            DiscountPolicy.title.ilike('%{query}%'.format(query=search)))
+
+    total_policies = discount_policies.count()
+    total_pages = int(math.ceil(total_policies / results_per_page))
+    offset = (page - 1) * results_per_page
+
+    discount_policies = discount_policies.limit(results_per_page).offset(offset).all()
```
jace: Use `query.paginate`. [It handles the details for you](http://flask-sqlalchemy.pocoo.org/2.1/api/#flask.ext.sqlalchemy.BaseQuery.paginate).

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,223 @@
...
+
+def make_discount(discount_policy, discount_policy_form, secret=False):
+    discount_policy_form.populate_obj(discount_policy)
+    if secret:
+        discount_policy.set_secret()
+    discount_policy.make_name()
+    return discount_policy
+
+
+@app.route('/admin/o/<org>/discount_policy/new', methods=['OPTIONS', 'POST'])
```
jace: Is handling `OPTIONS` necessary here? Don't we get this for free from Flask?

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,223 @@
...
+        discount_policies=discount_policies,
+        total_pages=total_pages,
+        paginated=(total_policies > results_per_page),
+        current_page=page)
+
+
+def make_discount(discount_policy, discount_policy_form, secret=False):
+    discount_policy_form.populate_obj(discount_policy)
+    if secret:
+        discount_policy.set_secret()
```
jace: This is a little confusing. Since `set_secret` also accepts a `secret` parameter, it took me a moment to realise that the boolean "secret" and string "secret" are distinct.

boxoffice/views/admin_discount.py
```diff
@@ -0,0 +1,223 @@
...
+def admin_new_coupon(discount_policy):
+    coupon_form = DiscountCouponForm(parent=discount_policy)
+    coupons = []
+    if not coupon_form.validate_on_submit():
+        return api_error(message=_(u"The coupon could not be created. Please rectify the indicated issues"),
+            status_code=400, errors=coupon_form.errors)
+    if coupon_form.count.data > 1:
+        # Create a signed discount coupon code
+        if not discount_policy.secret:
+            discount_policy.set_secret()
```
jace: As above.

boxoffice/views/admin_item.py
jace: Needs to be in the models as well.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,99 @@
...
+    discount_type = forms.RadioField(__("Discount type"),
+        choices=DISCOUNT_TYPE.items(), coerce=int, default=DISCOUNT_TYPE.COUPON)
+    is_price_based = forms.RadioField(__("Price based discount"),
+        coerce=getbool, default=True, choices=[
+            (True, __("Special price discount")),
+            (False, __("Percentage based discount"))])
+
+
+def validate_unique_discount_code_base(form, field):
+    if DiscountPolicy.query.filter(DiscountPolicy.id != form.edit_id, DiscountPolicy.discount_code_base == field.data).notempty():
```
shreyas-satish: Issued a PR - https://github.com/hasgeek/baseframe/pull/149/files

alembic/versions/36f458047cfd_update_index_on_discount_title.py
```diff
@@ -0,0 +1,25 @@
...
+down_revision = '3a585b8d5f8d'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+def upgrade():
+    op.drop_constraint(u'discount_policy_discount_code_base_key', 'discount_policy', type_='unique')
+    op.create_unique_constraint(u'discount_policy_organization_discount_code_base_key', 'discount_policy', ['organization_id', 'discount_code_base'])
```
shreyas-satish: Fixed.

alembic/versions/36f458047cfd_update_index_on_discount_title.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def downgrade():
+    op.drop_constraint('discount_policy_organization_discount_code_base_key', 'discount_policy', type_='unique')
+    op.create_unique_constraint(u'discount_policy_discount_code_base_key', 'discount_policy', ['discount_code_base'])
```
shreyas-satish: Fixed.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,99 @@
...
+class DiscountPolicyForm(forms.Form):
+    title = forms.StringField(__("Discount title"),
+        validators=[forms.validators.DataRequired(__("Please specify a discount title")),
+        forms.validators.Length(max=250)], filters=[forms.filters.strip()])
+    discount_type = forms.RadioField(__("Discount type"),
+        choices=DISCOUNT_TYPE.items(), coerce=int, default=DISCOUNT_TYPE.COUPON)
+    is_price_based = forms.RadioField(__("Price based discount"),
+        coerce=getbool, default=True, choices=[
+            (True, __("Special price discount")),
+            (False, __("Percentage based discount"))])
```
shreyas-satish: Fixed.

boxoffice/forms/discount.py
```diff
@@ -0,0 +1,99 @@
...
+            forms.validators.NumberRange(min=1, max=100, message=__("Percentage should be between >= 1 and <= 100"))])
+    discount_code_base = forms.StringField(__("Discount title"),
+        validators=[forms.validators.DataRequired(__("Please specify a discount code base")),
+        forms.validators.Length(max=20), validate_unique_discount_code_base],
+        filters=[forms.filters.strip(), forms.filters.none_if_empty()])
+    bulk_coupon_usage_limit = forms.IntegerField(__("Bulk coupon usage limit"), default=1)
+
+    def set_queries(self):
+        self.items.query = Item.query.join(ItemCollection).filter(
+            ItemCollection.organization == self.edit_parent).options(db.load_only('id', 'title'))
```
shreyas-satish: It's actually in a similar format i.e `item_collection_title: item_title`. `select2` has been configured to hit a different endpoint for the titles and ids - https://github.com/hasgeek/boxoffice/blob/admin-discount/boxoffice/views/admin_item.py#L28.

boxoffice/models/discount_policy.py
```diff
@@ -27,16 +28,18 @@ class DISCOUNT_TYPE(LabeledEnum):
 class DiscountPolicy(BaseScopedNameMixin, db.Model):
     """
     Consists of the discount rules applicable on items
+
+    `title` has a GIN index to enable trigram matching.
```
shreyas-satish: Fixed.

boxoffice/views/admin_discount.py
shreyas-satish: Changed to use `query.paginate`.

boxoffice/views/admin_discount.py
shreyas-satish: Yes, `OPTIONS` wasn't necessary, removed it here and from some of the other methods in this file.

boxoffice/templates/layout.html
vidya-ram: I have reverted the changes. Will a issue a new PR.

boxoffice/models/discount_policy.py
```diff
@@ -95,6 +98,11 @@ def get_from_signed_code(cls, code):
         except BadSignature:
             return None
 
+    def set_secret(self, secret=None):
```
shreyas-satish: Moved the secret setting to `__init__` and removed `set_secret`.

boxoffice/templates/index.html
jace: Let the move to Baseframe be in a separate PR.

boxoffice/views/admin_item.py
jace: This LIKE query is still here. Move to the index?

PR #139, opened 2017-03-23
Adds support for partial refunds for paid orders

shreyas-satish:


boxoffice/templates/order_refund_mail.html
```diff
@@ -0,0 +1,49 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset='UTF-8'>  
+  <title>Ticket Cancellation</title>
```
jace: Not a cancellation.

boxoffice/views/order.py
```diff
@@ -433,6 +433,53 @@ def cancel_line_item(line_item):
...
+
+        total_paid_amount = sum([order_transaction.amount
+            for order_transaction in order_payment_transactions])
+
+        total_refunded_amount = sum([order_transaction.amount
+            for order_transaction in order.get_refund_transactions()])
+
+        if (total_refunded_amount + requested_refund_amount) >= total_paid_amount:
+            return make_response(jsonify(status='error', error='excess_partial_refund',
+                error_description='Invalid refund amount, must be lesser than net amount paid for the order'), 403)
```
jace: This validation should be in the form.

jace: Looks like the `from_json` constructor doesn't take the usual init parameters, which is why these validations can't be moved into the form (form doesn't have `edit_obj` set). This is a separate ticket to fix now.

boxoffice/templates/order_refund_mail.html
shreyas-satish: Fixed.

boxoffice/templates/order_refund_mail.html
```diff
@@ -0,0 +1,49 @@
...
+  </style>
+</head>
+
+<body>
+  <div>
+
+    <p>Hello {{ order.buyer_fullname }},</p>
+
+    <p>
+      A refund of {{ currency_symbol }} {{ refund_amount }} has been initiated against your order (Receipt No: {{ order.invoice_no }}). You should receive the refund in 2-5 days.</p>
```
shreyas-satish: Why is the user getting a refund? It would help if the admin who is issuing a refund, can leave two notes:

1. An internal note that explains why this refund was issued.
2. A note to the user to explain why they're getting a refund.

boxoffice/views/order.py
```diff
@@ -433,6 +433,55 @@ def cancel_line_item(line_item):
...
+            pg_payment_status=RAZORPAY_PAYMENT_STATUS.CAPTURED).one()
+        rp_resp = razorpay.refund_payment(payment.pg_paymentid, requested_refund_amount)
+        if rp_resp.status_code == 200:
+            transaction = PaymentTransaction(order=order, transaction_type=TRANSACTION_TYPE.REFUND,
+                online_payment=payment, amount=requested_refund_amount, currency=CURRENCY.INR,
+                internal_note=form.internal_note.data, note_to_user=form.note_to_user.data)
+            db.session.add(transaction)
+            db.session.commit()
+            send_order_refund_mail.delay(order.id, transaction.amount, transaction.note_to_user)
+            return make_response(jsonify(status='ok', result={'message': 'Refund processed for order'}), 200)
```
vidya-ram: Better to send the latest order amount to the client so that the UI could be updated.

shreyas-satish: Very good catch, thank you for pointing this out.

boxoffice/forms/order.py
```diff
@@ -55,3 +56,12 @@ class OrderSessionForm(forms.Form):
...
     referrer = forms.StringField(__("Referrer"), filters=[trim(2083)])
+
+
+class OrderRefundForm(forms.Form):
+    amount = forms.IntegerField(__("Amount"),
+        validators=[forms.validators.DataRequired(__("Please specify an amount"))])
+    internal_note = forms.TextAreaField(__("Internal note"),
+        description=__("Add a note for future reference"))
+    note_to_user = forms.TextAreaField(__("Note to user"),
+        description=__("Leave a note for the buyer"))
```
jace: "Send this note to the buyer"

boxoffice/mailclient.py
```diff
@@ -58,6 +59,24 @@ def send_line_item_cancellation_mail(line_item_id, refund_amount, subject="Ticke
 
 
 @job('boxoffice')
+def send_order_refund_mail(order_id, refund_amount, note_to_user):
+    with app.test_request_context():
+        order = Order.query.get(order_id)
+        subject = __("{item_collection_title}: Order refund for receipt no: {invoice_no}".format(
```
jace: "{item_collection_title}: Refund for receipt no. {invoice_no}" (no "order" and a period)

boxoffice/models/payment.py
```diff
@@ -73,6 +69,21 @@ class PaymentTransaction(BaseMixin, db.Model):
     transaction_method = db.Column(db.Integer, default=TRANSACTION_METHOD.ONLINE, nullable=False)
     # Eg: reference number for a bank transfer
     transaction_ref = db.Column(db.Unicode(80), nullable=True)
+    refunded_at = db.Column(db.DateTime, nullable=True)
+    internal_note = db.Column(db.Text, default=u'', nullable=True)
+    note_to_user = db.Column(db.Text, default=u'', nullable=True)
```
jace: Both should have no default. They're nullable.

boxoffice/forms/order.py
jace: If these are optional, add `filters=[null_if_blank()]`

boxoffice/templates/order_refund_mail.html
```diff
@@ -0,0 +1,53 @@
...
+<body>
+  <div>
+
+    <p>Hello {{ order.buyer_fullname }},</p>
+
+    <p>
+      A refund of {{ currency_symbol }} {{ refund_amount }} has been initiated against your order (Receipt No: {{ order.invoice_no }}). You should receive the refund in 2-5 days.
+    </p>
+    <p>
+      {{note_to_user}}
```
jace: What happens if the text contains newlines? It's a textarea widget in the form. When we write an explanation to the user, it will have multiple lines.

boxoffice/views/order.py
jace: As these error messages are human readable strings, use double quotes (the one above as well).

boxoffice/views/order.py
```diff
@@ -433,6 +434,56 @@ def cancel_line_item(line_item):
...
+        else:
+            raise PaymentGatewayError("Refund failed for order - {order} with the following details - {msg}".format(order=order.id,
+                msg=rp_resp.content), 424,
+            'Refund failed. Please try again or contact support at {email}.'.format(email=order.organization.contact_email))
+    else:
+        return make_response(jsonify(status='error', error='invalid_input',
+            error_description='Please enter a valid amount and currency'), 403)
```
jace: Double quoted error descriptions.

boxoffice/templates/order_refund_mail.html
shreyas-satish: Switched to a markdown column.

boxoffice/forms/order.py
```diff
@@ -55,3 +55,14 @@ class OrderSessionForm(forms.Form):
...
     referrer = forms.StringField(__("Referrer"), filters=[trim(2083)])
+
+
+class OrderRefundForm(forms.Form):
+    amount = forms.IntegerField(__("Amount"),
+        validators=[forms.validators.DataRequired(__("Please specify an amount"))])
+    internal_note = forms.TextAreaField(__("Internal note"),
+        description=__("Add a note for future reference"), filters=[forms.filters.none_if_empty()])
+    refund_description = forms.TextAreaField(__("Refund description"),
+        description=__("Why is this order receiving a refund?"), filters=[forms.filters.none_if_empty()])
```
jace: This should be a text field. We're not handling newlines here either.

boxoffice/forms/order.py
```diff
@@ -55,3 +55,14 @@ class OrderSessionForm(forms.Form):
     utm_id = forms.StringField(__("UTM Id"), filters=[trim(250)])
     gclid = forms.StringField(__("Gclid"), filters=[trim(250)])
     referrer = forms.StringField(__("Referrer"), filters=[trim(2083)])
+
+
+class OrderRefundForm(forms.Form):
+    amount = forms.IntegerField(__("Amount"),
+        validators=[forms.validators.DataRequired(__("Please specify an amount"))])
+    internal_note = forms.TextAreaField(__("Internal note"),
+        description=__("Add a note for future reference"), filters=[forms.filters.none_if_empty()])
```
jace: As below, just a string field.

alembic/versions/18576fdffd86_add_notes_and_refund_timestamp_to_.py
```diff
@@ -0,0 +1,31 @@
...
+down_revision = '4246213b032b'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+def upgrade():
+    op.add_column('payment_transaction', sa.Column('internal_note', sa.Text(), nullable=True))
+    op.add_column('payment_transaction', sa.Column('refund_description', sa.Text(), nullable=True))
```
jace: These two should be regular string fields, not text areas, unless we're specifying a rendering format.

boxoffice/forms/order.py
```diff
@@ -3,7 +3,7 @@
 from baseframe import __
 import baseframe.forms as forms
 
-__all__ = ['LineItemForm', 'BuyerForm', 'OrderSessionForm']
+__all__ = ['LineItemForm', 'BuyerForm', 'OrderSessionForm', 'OrderRefundForm']
 
 
 def trim(length):
```
jace: Unrelated, but can we move `trim` into `baseframe.forms.filters` now that that module exists?

boxoffice/views/order.py
```diff
@@ -433,6 +435,52 @@ def cancel_line_item(line_item):
...
+                error_description='Invalid refund amount, must be lesser than net amount paid for the order'), 403)
+
+        payment = OnlinePayment.query.filter_by(order=order,
+            pg_payment_status=RAZORPAY_PAYMENT_STATUS.CAPTURED).one()
+        rp_resp = razorpay.refund_payment(payment.pg_paymentid, requested_refund_amount)
+        if rp_resp.status_code == 200:
+            transaction = PaymentTransaction(order=order, transaction_type=TRANSACTION_TYPE.REFUND,
+                online_payment=payment, amount=requested_refund_amount, currency=CURRENCY.INR,
+                refunded_at=func.utcnow(), refund_description=form.refund_description.data,
+                internal_note=form.internal_note.data, note_to_user=form.note_to_user.data)
```
jace: You should ideally be using `form.populate_obj(transaction)`. Right now you're hardcoding the fields you want, which is splitting logic between the form and the view.

boxoffice/mailclient.py
```diff
@@ -58,6 +59,24 @@ def send_line_item_cancellation_mail(line_item_id, refund_amount, subject="Ticke
 
 
 @job('boxoffice')
+def send_order_refund_mail(order_id, refund_amount, note_to_user):
+    with app.test_request_context():
+        order = Order.query.get(order_id)
+        subject = __("{item_collection_title}: Refund for receipt no. {invoice_no}".format(
```
jace: I think the subject should say "partial refund" to avoid confusing users.

boxoffice/models/payment.py
```diff
@@ -73,6 +69,40 @@ class PaymentTransaction(BaseMixin, db.Model):
     transaction_method = db.Column(db.Integer, default=TRANSACTION_METHOD.ONLINE, nullable=False)
     # Eg: reference number for a bank transfer
     transaction_ref = db.Column(db.Unicode(80), nullable=True)
+    refunded_at = db.Column(db.DateTime, nullable=True)
+    internal_note = db.Column(db.Text, nullable=True)
+    refund_description = db.Column(db.Text, nullable=True)
```
jace: These two should be `Unicode(250)`

boxoffice/forms/order.py
jace: A textarea field works (for the sake of rendering long strings) as long as we strip newlines from it. Hasjob does this in the jobpost reject form, for instance.

boxoffice/templates/order_refund_mail.html
```diff
@@ -0,0 +1,51 @@
...
+  <div>
+
+    <p>Hello {{ order.buyer_fullname }},</p>
+
+    <p>
+      A refund of {{ currency_symbol }} {{ refund_amount }} has been initiated against your order (Receipt No: {{ order.invoice_no }}). You should receive the refund in 2-5 days.
+    </p>
+    {{note_to_user}}
+    <p class="footer">Thank you,</p>
+    <p>{{ org.title }}</p>
```
jace: Don't hardcode this in the template. Let it be in the user-provided note.

boxoffice/mailclient.py
shreyas-satish: There might be a case where the user gets a full refund (eg: ticket buyer who gets selected as a speaker), but whose ticket stays confirmed. The body of the email which explains the amount refunded should clarify this?

PR #143, opened 2017-04-10
Fix cash receipt for cancelled and discounted tickets.

vidya-ram:


boxoffice/templates/cash_receipt.html
```diff
@@ -212,7 +212,7 @@ <h1 class="header">Cash Receipt</h1>
       
       {% for line_item in line_items -%}
         <tr>
-          <td class="center" rowspan="2" valign="top">{{ line_item.line_item_seq }}</td>
+          <td class="center" rowspan="2" valign="top">{{ loop.index }}</td>
```
shreyas-satish: Why are we using `loop.index` instead of `line_item.line_item_seq`?

vidya-ram: We are using `loop.index` to indicate the row number of the table.
`line_item.line_item_seq` will not be in sequence incase we are cancelling bulk discounted tickets.

PR #158, opened 2017-06-05
Redesign ticket in the email confirmation and UI changes to boxoffice confirm screen.

vidya-ram:
1. Confirmation screen UI changes on Boxoffice widget:
screen shot 2017-06-05 at 11 28 38 am

2. Ticket redesign on the email confirmation
screen shot 2017-06-05 at 11 26 47 am

### DIFF
--- a/boxoffice/templates/boxoffice.html
+++ b/boxoffice/templates/boxoffice.html
@@ -172,31 +172,52 @@
     {{elseif id === 'boxoffice-confirm' && id === activeTab}}
       <div id='{{id}}' class="boxoffice-section clearfix" intro='fly:{"x":600,"y":"0"}'>
         {{#section}}
-          <p class="confirmation-icon"><i class="fa fa-check-circle"></i></p>
-          <p class="confirmation-header">Thank you for your order!</p>
-          <div class="confirmation-msg">
-            <p class="">We’ve mailed you the cash receipt.</p>
-            <div class="receipt-btn-wrapper">
-              <a href="{{ cashReceiptURL }}" target="_blank" class="boxoffice-button boxoffice-button-action receipt-button" on-click="fire('eventAnalytics','view cash receipt', 'cash receipt')">See cash receipt</a>
-              <a href="{{ attendeeAssignmentURL }}" target="_blank" class="boxoffice-button boxoffice-button-action receipt-button" on-click="fire('eventAnalytics','fill attendee details', 'attendee details')">Fill attendee details</a>
-            </div>
-            <div class="social-icons">
-              <p>Tell your friends!</p>
-              <span class="tweet-button">
-              <a href="https://twitter.com/share" class="twitter-share-button" data-text="Yay! I'm attending {{eventTitle}}" data-hashtags="HasGeek{{#if eventHashtag}}, {{eventHashtag}}{{/if}}">Tweet</a>
-              </span>
-              <div class="fb-share-button pad-half" data-layout="button"></div>
-              <div id="fb-root"></div>
-              <script>(function(d, s, id) { var js, fjs = d.getElementsByTagName(s)[0]; if (d.getElementById(id)) return; js = d.createElement(s); js.id = id; js.src = "//connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.5"; fjs.parentNode.insertBefore(js, fjs); }(document, 'script', 'facebook-jssdk'));</script>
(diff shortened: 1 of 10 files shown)


shreyas-satish:
I suggest changing the copy in the first screenshot to the following format:

## Almost done!

Thank you for your order! You should receive a receipt in your mailbox soon. Meanwhile, care to help us with some more details? We want to make a nice badge for you!

The last sentence should ideally be dependent on the number of tickets that the buyer has bought. If it's more than one, it should be 'We want to make nice badges for you'.

tests/test_item_collection.py
```diff
@@ -7,7 +7,7 @@
 
 class TestItemCollectionAPI(unittest.TestCase):
 
-    expected_keys = ['categories', 'html', 'refund_policy']
+    expected_keys = ['categories', 'html', 'eventTitle', 'eventDetails', 'refund_policy']
```
shreyas-satish: This should probably be in snake_case.

boxoffice/templates/ticket_assignment_mail.html
```diff
@@ -105,25 +127,64 @@
...
+          {%- if item_collection.details.logo -%}
+            <img src="{{ item_collection.details.logo }}" class="logo"/>
+          {%- elif org.details.logo %}
+            <img src="{{ org.details.logo }}" class="logo"/>
+          {%- endif %}
+          <div class="event">
+            <p class="event-title">{{ item_collection.title }}</p>
+            <p class="event-date">
+            {%- if line_item.item.details.date -%}
+              <span>{{ line_item.item.details.date }}</span>
```
shreyas-satish: This is a slightly inefficient way to access data via a relationship. Since this doesn't impact user experience it's probably fine, but ideally you might want to fetch all the data you need with a join in the view, and pass the data to the template.

boxoffice/templates/boxoffice.html
```diff
@@ -172,31 +172,52 @@ <h1 class="category-heading">{{title}}</h1>
...
+                    </div>
+                  </div>
+                  <div class="ticket-body">
+                    <p class="ticket-form-label">Name</p>
+                    <p class="ticket-input"><span class="cursor">&#124;</span></p>
+                    <p class="ticket-form-label">Company</p>
+                    <p class="ticket-input"></p>
+                    {{#if getConfirmedTicket(order.line_items).item_details }}
+                      <p class="ticket-form-label">Venue</p>
+                      <p class="ticket-field-value">{{ getConfirmedTicket(order.line_items).item_details.venue }}</p>
```
shreyas-satish: Can we use a computed property instead of a function call?

boxoffice/templates/ticket_assignment_mail.html
```diff
@@ -105,25 +127,64 @@
...
+            {%- endif %}
+          </p>
+        </div>
+      </div>
+      <p><a class="edit-button" href={{"{base_url}/order/{access_token}/ticket".format(base_url=base_url, access_token=order.access_token)}} target="_blank">Edit details</a></p>
     </div>
 
-    <p><a class="button" href={{"{base_url}/order/{access_token}/ticket".format(base_url=base_url, access_token=order.access_token)}} target="_blank">Edit details</a></p>
-    <div class="terms-content">{{org.details.get('ticket_faq', '') | safe}}</div>
+    <div class="terms-content">{{ org.details.get('ticket_faq') | safe }}</div>
```
jace: This template has many hardcoded assumptions on the fields in the `details` JSON dictionary on various models. These fields need to be reflected in the data model itself, as there is no way to know what fields are expected without reading each line of the templates.

migrations/versions/3d55fce529de_add_details_to_item_itemcollection.py
```diff
@@ -0,0 +1,25 @@
...
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '3d55fce529de'
+down_revision = '36f458047cfd'
+
+from alembic import op
+import sqlalchemy as sa
+from coaster import sqlalchemy
```
jace: These two import invocations mean different things:
```python
from coaster import sqlalchemy
import coaster.sqlalchemy as sqlalchemy
```
For the first version to work, coaster's `__init__.py` must have an `import sqlalchemy`. This is obviously not a good idea. For `JsonDict`, always use `from coaster.sqlalchemy import JsonDict`

boxoffice/templates/ticket_assignment_mail.html
shreyas-satish: I think this points to a requirement of modeling the event i.e a domain model (say `Event`) that comprises of details such as `venue`, `start_at`, `end_at` that shares a many-to-many relationship with `Item` (to be renamed to `Ticket`). A many-to-many relationship is required to allow for the use case of mapping a combo-ticket to multiple events.

PR #164, opened 2017-07-04
Adds models and workflow for capturing invoice details

shreyas-satish:


boxoffice/models/invoice.py
```diff
@@ -0,0 +1,55 @@
...
+    invoiced_at = db.Column(db.DateTime, nullable=True)
+    street_address = db.Column(db.Unicode(255), nullable=True)
+    city = db.Column(db.Unicode(255), nullable=True)
+    state = db.Column(db.Unicode(255), nullable=True)
+    country = db.Column(db.Unicode(255), nullable=True)
+    postcode = db.Column(db.Unicode(8), nullable=True)
+    # GSTIN in the case of India
+    taxid = db.Column(db.Unicode(255), nullable=True)
+    # India specific: this is the state short code. Eg: KA for Karnataka
+    place_of_supply = db.Column(db.Unicode(3), nullable=True)
```
jace: Replace this with `state_code`.

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,55 @@
...
+    __table_args__ = (db.UniqueConstraint('organization_id', 'invoice_no'),)
+
+    invoicee_name = db.Column(db.Unicode(255), nullable=True)
+    invoicee_email = db.Column(db.Unicode(254), nullable=True)
+    invoice_no = db.Column(db.Unicode(32), nullable=True)
+    invoiced_at = db.Column(db.DateTime, nullable=True)
+    street_address = db.Column(db.Unicode(255), nullable=True)
+    city = db.Column(db.Unicode(255), nullable=True)
+    state = db.Column(db.Unicode(255), nullable=True)
+    country = db.Column(db.Unicode(255), nullable=True)
```
jace: `country` should be an ISO country code. No need to store names. 2 characters.

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,55 @@
...
+    postcode = db.Column(db.Unicode(8), nullable=True)
+    # GSTIN in the case of India
+    taxid = db.Column(db.Unicode(255), nullable=True)
+    # India specific: this is the state short code. Eg: KA for Karnataka
+    place_of_supply = db.Column(db.Unicode(3), nullable=True)
+
+    customer_order_id = db.Column(None, db.ForeignKey('customer_order.id'), nullable=False, index=True, unique=False)
```
jace: `unique=False` isn't necessary.

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,55 @@
...
+    order = db.relationship('Order', backref=db.backref('invoices', cascade='all, delete-orphan'))
+
+    organization_id = db.Column(None, db.ForeignKey('organization.id'), nullable=False)
+    organization = db.relationship('Organization', backref=db.backref('invoices', cascade='all, delete-orphan', lazy='dynamic'))
```
jace: If orders also belong to an organization, this is denormalizing it.  You'll need a column validator to ensure this field can never point to another organization. (I can understand denormalization as an indexing strategy, but this sort of thing can result in pretty bad data integrity errors.)

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,55 @@
...
+    gst_type = db.Column(db.Unicode(7), nullable=False)
+    discount_title = db.Column(db.Unicode(255), nullable=False)
+
+    currency = db.Column(db.Unicode(3), nullable=False)
+    base_amount = db.Column(db.Numeric, default=Decimal(0), nullable=False)
+    discounted_amount = db.Column(db.Numeric, default=Decimal(0), nullable=False)
+    final_amount = db.Column(db.Numeric, default=Decimal(0), nullable=False)
+
+    invoice_id = db.Column(None, db.ForeignKey('invoice.id'), nullable=False, index=True, unique=False)
+    invoice = db.relationship(Invoice, backref=db.backref('line_items', cascade='all, delete-orphan'))
```
jace: Use an [ordering list](http://docs.sqlalchemy.org/en/latest/orm/extensions/orderinglist.html) so `seq` is auto-maintained.

boxoffice/models/item.py
```diff
@@ -25,6 +32,7 @@ class Item(BaseScopedNameMixin, db.Model):
             collection_class=ordering_list('seq', count_from=1)))
 
     parent = db.synonym('item_collection')
+    gst_type = db.Column(db.Integer, default=GST_TYPE.SERVICE, nullable=True)
```
jace: `db.SmallInteger`

boxoffice/models/item_collection.py
```diff
@@ -18,3 +18,4 @@ class ItemCollection(BaseScopedNameMixin, db.Model):
     organization_id = db.Column(db.Integer, db.ForeignKey('organization.id'), nullable=False)
     organization = db.relationship(Organization, backref=db.backref('item_collections', cascade='all, delete-orphan'))
     parent = db.synonym('organization')
+    tax_type = db.Column(db.Unicode(80), nullable=True, default='GST')
```
jace: `u'GST'`

boxoffice/models/order.py
```diff
@@ -62,6 +62,7 @@ class Order(BaseMixin, db.Model):
     buyer_fullname = db.Column(db.Unicode(80), nullable=False)
     buyer_phone = db.Column(db.Unicode(16), nullable=False)
 
+    # TODO: Rename to receipt_no
     invoice_no = db.Column(db.Integer, nullable=True)
```
jace: Shouldn't this have a foreign key reference?

boxoffice/models/invoice.py
jace: Use ISO 3166-1 and 3166-2 codes for both country and state.

Also note: India has three ways to represent state codes. [ISO-3166-2:IN](https://en.wikipedia.org/wiki/ISO_3166-2:IN) defines ISO state codes, but some states have different codes for license plates. Uttarakhand (formerly Uttaranchal) is variously `UT`, `UK`, `UA` and `UL`. The tax codes, which are numeric, are a [different series](http://www.caa-ahm.org/Pdf/Legal/Legal-35.pdf).

boxoffice/models/order.py
shreyas-satish: `Invoice` has a foreign key reference to `Order`.

jace: Foreign key constraint.

jace: Unless this is an actual number and not a reference. If so, why integer?

shreyas-satish: This is an actual number that is incremented for every order within the scope of an organization.

jace: Why are we not using Coaster's existing `BaseScopedIdMixin` then? It includes a safety harness (server-side increments) learnt from failures in production.

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,93 @@
...
+    __tablename__ = 'invoice_line_item'
+    __uuid_primary_key__ = True
+    __table_args__ = (db.UniqueConstraint('invoice_id', 'seq'),)
+
+    seq = db.Column(db.Integer, nullable=False)
+    item_title = db.Column(db.Unicode(255), nullable=False)
+    quantity = db.Column(db.SmallInteger, nullable=False, default=1)
+    # In India, this will be GST
+    tax_type = db.Column(db.Unicode(255), nullable=False)
+    gst_type = db.Column(db.Unicode(7), nullable=False)
```
jace: What is the GST type if tax type is not GST? Surely this needs `nullable=True`?

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class InvoiceLineItem(UuidMixin, BaseMixin, db.Model):
+    __tablename__ = 'invoice_line_item'
+    __uuid_primary_key__ = True
+    __table_args__ = (db.UniqueConstraint('invoice_id', 'seq'),)
+
+    seq = db.Column(db.Integer, nullable=False)
+    item_title = db.Column(db.Unicode(255), nullable=False)
+    quantity = db.Column(db.SmallInteger, nullable=False, default=1)
```
jace: `SmallInteger` is typically warranted when you have an enumeration, like status numbers.

boxoffice/models/invoice.py
jace: Perhaps the field should be called `tax_subtype` for flexibility in other tax regimes.

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,93 @@
...
+    currency = db.Column(db.Unicode(3), nullable=False)
+    base_amount = db.Column(db.Numeric, default=Decimal(0), nullable=False)
+    discounted_amount = db.Column(db.Numeric, default=Decimal(0), nullable=False)
+    # Total amount includes tax
+    total_amount = db.Column(db.Numeric, default=Decimal(0), nullable=False)
+
+    cgst_tax_rate = db.Column(db.Integer, nullable=True, default=0)
+    sgst_tax_rate = db.Column(db.Integer, nullable=True, default=0)
+    igst_tax_rate = db.Column(db.Integer, nullable=True, default=0)
+    gst_compensation_cess = db.Column(db.SmallInteger, nullable=True, default=0)
```
jace: Are we hardcoding for all GST subtypes here? What about `UTGST`? Does it need to be separate, or is it equivalent to `SGST` (in filings)? If this is the rate stored in each line item, what is the actual value? Considering it will not always round-off to two digits when you calculate from the base rate, shouldn't you be storing the actual rounded-off value?

boxoffice/models/item.py
```diff
@@ -26,6 +33,19 @@ class Item(BaseScopedNameMixin, db.Model):
...
+    hsn = db.Column(db.Unicode(255), nullable=True)
+    # Services Accounting Code
+    sac = db.Column(db.Unicode(255), nullable=True)
+    cgst_tax_rate = db.Column(db.SmallInteger, nullable=True, default=0)
+    sgst_tax_rate = db.Column(db.SmallInteger, nullable=True, default=0)
+    igst_tax_rate = db.Column(db.SmallInteger, nullable=True, default=0)
+    gst_compensation_cess = db.Column(db.SmallInteger, nullable=True, default=0)
+    state_code = db.Column(db.Unicode(4), nullable=True)
+    # ISO country code
+    country_code = db.Column(db.Unicode(2), nullable=True)
```
jace: We're storing all of these fields on the `Item` model? What happens in the following scenario?

1. We're selling an item (at multiple price points) from July 1 to November 30 for an event on December 1.
2. On July 31, the government announces that the GST rate for event tickets will change from 18% back to 15% effective September 1.
3. We have the month of August to implement a rate change starting from September. Where is this change recorded in our software? If you edit the Item record's rates, it starts applying immediately. Do you set a reminder to edit it on the midnight of September 1?

boxoffice/models/user.py
```diff
@@ -28,10 +28,12 @@ class Organization(ProfileBase, db.Model):
     __tablename__ = 'organization'
     __table_args__ = (db.UniqueConstraint('contact_email'),)
 
-    # The currently used fields in details are address(html), cin (Corporate Identity Number), pan, service_tax_no, support_email,
+    # The currently used fields in details are address(html), cin (Corporate Identity Number)
+    # pan service_tax_no, support_email,
     # logo (image url), refund_policy (html), ticket_faq (html), website (url)
     details = db.Column(JsonDict, nullable=False, server_default='{}')
     contact_email = db.Column(db.Unicode(254), nullable=False)
+    fy_start_at = db.Column(db.DateTime, nullable=True)
```
jace: As above. This will get you into trouble on April 1 should this be forgotten.

boxoffice/views/item_collection.py
```diff
@@ -50,7 +52,9 @@ def jsonify_category(category):
 def boxofficejs():
     return make_response(jsonify({
         'script': render_template('boxoffice.js', base_url=request.url_root.strip('/'),
-        razorpay_key_id=app.config['RAZORPAY_KEY_ID'])
+        razorpay_key_id=app.config['RAZORPAY_KEY_ID'],
+        states=[{'name': state['name'], 'code': state['short_code_text']} for state in indian_states],
+        countries=[{'name': country.name, 'code': country.alpha_2} for country in pycountry.countries])
```
jace: Both lists need to be sorted.

boxoffice/views/order.py
```diff
@@ -269,7 +273,7 @@ def payment(order):
                 db.session.add(line_item.discount_coupon)
         db.session.commit()
         send_receipt_mail.delay(order.id)
-        return make_response(jsonify(message="Payment verified"), 201)
+        return make_response(jsonify(invoice_id=invoice.id), 201)
```
jace: Use a standard format for JSON responses. We agreed on this for API responses:
```json
{
    "status": "ok",
    "result": {
        "invoice_id": 0
    }
}
```

jace: In fact, use the `api_success` function you already have.

boxoffice/views/utils.py
```diff
@@ -56,7 +73,14 @@ def add_headers(resp, origin):
     @wraps(f)
     def wrapper(*args, **kwargs):
         origin = request.headers.get('Origin')
-        if not origin or origin not in app.config['ALLOWED_ORIGINS']:
+        if not origin:
+            # Firefox doesn't send the Origin header, so read the Referer header instead
+            # TODO: Remove this conditional when Firefox starts adding an Origin header
+            referer = request.headers.get('Referer')
```
jace: Use `request.referrer`

boxoffice/forms/order.py
```diff
@@ -68,3 +69,33 @@ class RefundTransactionForm(forms.Form):
         description=__("Why is this order receiving a refund?"), filters=[forms.filters.none_if_empty()])
     note_to_user = forms.MarkdownField(__("Note to user"),
         description=__("Send this note to the buyer"), filters=[forms.filters.none_if_empty()])
+
+
+def validate_state_code(form, field):
+    # Note: state_code is only a required field if the chosen country is India
+    if form.country_code.data == "IN":
+        if field.data not in [state['short_code_text'] for state in indian_states]:
+            raise forms.validators.StopValidation(__("Please select a valid Indian state"))
```
vidya-ram: Fixed

boxoffice/forms/order.py
```diff
@@ -68,3 +69,33 @@ class RefundTransactionForm(forms.Form):
...
+        forms.validators.Length(max=255)], filters=[forms.filters.strip()])
+    city = forms.StringField(__("City"), validators=[forms.validators.DataRequired(__("Please enter the city")),
+        forms.validators.Length(max=255)], filters=[forms.filters.strip()])
+    country_code = forms.StringField(__("Country"), validators=[forms.validators.DataRequired(__("Please select a country")),
+        forms.validators.Length(max=2)], filters=[forms.filters.strip()])
+    state_code = forms.StringField(__("State code"), validators=[forms.validators.Length(max=4),
+        validate_state_code], filters=[forms.filters.strip()])
+    state = forms.StringField(__("State"), validators=[forms.validators.Optional(),
+        forms.validators.Length(max=255)], filters=[forms.filters.strip(), forms.filters.none_if_empty()])
+    postcode = forms.StringField(__("Pincode"), validators=[forms.validators.DataRequired(__("Please enter a postcode")),
```
vidya-ram: Fixed

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,93 @@
...
+from sqlalchemy import event
+from sqlalchemy import sql, func
+from sqlalchemy.ext.orderinglist import ordering_list
+
+__all__ = ['Invoice', 'InvoiceLineItem']
+
+
+def get_latest_invoice_no(organization):
+    """
+    Returns the last invoice number used, 0 if no order has ben invoiced yet
```
vidya-ram: Fixed

boxoffice/views/utils.py
vidya-ram: Fixed

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    query = db.session.query(sql.functions.max(Invoice.invoice_no))\
+        .filter(Invoice.organization == organization)
+    if organization.fy_start_at:
```
shreyas-satish: Here's an alternate approach: we store `fy_start_month` and `fy_start_day`, and dynamically compute `fy_start_at`.

boxoffice/models/item.py
shreyas-satish: Removing these fields, and making a note of this comment for the next iteration.

boxoffice/models/invoice.py
shreyas-satish: `fy_start_at` is now computed dynamicall based on `fy_start_month`, `fy_start_day` and `fy_timezone`.

boxoffice/models/invoice.py
shreyas-satish: Fixed, and changed to `tax_subtype`.

boxoffice/models/invoice.py
shreyas-satish: Changed to `Integer`.

boxoffice/views/order.py
vidya-ram: Fixed

boxoffice/models/user.py
```diff
@@ -24,14 +27,52 @@ def default_user(context):
...
+        db.CheckConstraint('fy_start_day >= 1 and fy_start_month <= 31', 'org_day_check'))
 
-    # The currently used fields in details are address(html), cin (Corporate Identity Number), pan, service_tax_no, support_email,
+    # The currently used fields in details are address(html), cin (Corporate Identity Number)
+    # pan service_tax_no, support_email,
     # logo (image url), refund_policy (html), ticket_faq (html), website (url)
     details = db.Column(JsonDict, nullable=False, server_default='{}')
     contact_email = db.Column(db.Unicode(254), nullable=False)
+    fy_start_month = db.Column(db.Integer, nullable=True)
+    fy_start_day = db.Column(db.Integer, nullable=True)
```
jace: What happens if dear government changes the financial year start from Apr 1 to Jan 1, as is widely rumoured? Editing these fields will now break past calculations.

shreyas-satish: If it's going to be changed to Jan 1, I suppose this will be announced _before_ Jan 1? In which case, we'll just need to update it on the 31st night, and `fy_start_at` will insure that invoice numbers will start over from 1.

Also, since this is an exception case, we'll need to manually intervene. But, apart from this case, `fy_start_at` will ensure the correct datetime every year, once `fy_start_month` and `fy_start_day` are set.

jace: Let's assume the last invoice number in the last FY was 1000 and the last invoice in the current FY is 100. You're running a query for invoices > start_date. If the government announces a date change to Jan 1 and you update it in the database, your query changes to look for the largest invoice number since Jan 1 of this year instead of Apr 1 (as the change becomes effective only _next_ year) and you'll incorrectly generate 1001 instead of 101. You should be recording when the last FY ended, not when the current FY started.

I suggest stepping back and thinking about this a bit. Look at how pytz handles historical time zone changes.

shreyas-satish: > If the government announces a date change to Jan 1 and you update it in the database, your query changes to look for the largest invoice number since Jan 1 of this year instead of Apr 1 (as the change becomes effective only next year) and you'll incorrectly generate 1001 instead of 101.

The plan was to update the database with the new `fy_start_month` and `fy_start_day` on the 1st midnight, in which case the invoice numbers for orders post the 1st midnight will start from 1.

> I suggest stepping back and thinking about this a bit. Look at how pytz handles historical time zone changes.

Sure, I'm rethinking about this to see if this can be done better.

boxoffice/forms/order.py
```diff
@@ -68,3 +69,33 @@ class RefundTransactionForm(forms.Form):
         description=__("Why is this order receiving a refund?"), filters=[forms.filters.none_if_empty()])
     note_to_user = forms.MarkdownField(__("Note to user"),
         description=__("Send this note to the buyer"), filters=[forms.filters.none_if_empty()])
+
+
+def validate_state_code(form, field):
+    # Note: state_code is only a required field if the chosen country is India
+    if form.country_code.data == "IN":
+        if field.data not in [state['short_code_text'] for state in indian_states]:
```
vidya-ram: Fixed

boxoffice/views/item_collection.py
vidya-ram: Fixed

boxoffice/models/user.py
jace: You need a FY lookup function that, given an invoice date and a jurisdiction, returns the FY boundaries (start and end). This of course gets complicated in the US where the default is Jan 1 - Dec 31, _but_ companies can choose their own FY.

shreyas-satish: Added a `get_financial_year` function.

boxoffice/models/invoice.py
shreyas-satish: > What about UTGST? Does it need to be separate, or is it equivalent to SGST (in filings)? 

They are equivalent.

> If this is the rate stored in each line item, what is the actual value? 

The actual value is computed when required.

> Considering it will not always round-off to two digits when you calculate from the base rate, shouldn't you be storing the actual rounded-off value?

I chose not to store actual values, since calculating the tax amount based on a constant tax rate and a constant base rate will always result in the same value.

I'm going to remove the `InvoiceLineItem` model for now and add it back in the next release after iterating on the model design.

boxoffice/forms/order.py
```diff
@@ -68,3 +69,33 @@ class RefundTransactionForm(forms.Form):
...
+            raise forms.validators.StopValidation(__("Please select a valid Indian state"))
+
+
+class InvoiceForm(forms.Form):
+    buyer_taxid = forms.StringField(__("GSTIN"), validators=[forms.validators.Optional(),
+        forms.validators.Length(max=255)], filters=[forms.filters.strip(), forms.filters.none_if_empty()])
```
shreyas-satish: Added a `GSTIN` validator.

boxoffice/forms/order.py
```diff
@@ -41,9 +42,9 @@ def process_list(cls, line_items_json):
 
 
 class BuyerForm(forms.Form):
-    email = forms.EmailField(__("Email"), validators=[forms.validators.DataRequired(), forms.validators.Length(max=80)])
-    fullname = forms.StringField(__("Full name"), validators=[forms.validators.DataRequired(), forms.validators.Length(max=80)])
-    phone = forms.StringField(__("Phone number"), validators=[forms.validators.Length(max=16)])
+    email = forms.EmailField(__("Email"), validators=[forms.validators.DataRequired(__("Please enter an email address")), forms.validators.Length(min=5, max=80)])
+    fullname = forms.StringField(__("Full name"), validators=[forms.validators.DataRequired(__("Please enter the buyer's fullname")), forms.validators.Length(max=80)])
```
jace: "full name" with a space.

boxoffice/forms/order.py
```diff
@@ -68,3 +69,52 @@ class RefundTransactionForm(forms.Form):
...
+    - 15 characters in length
+    - First two characters form a valid short code for an Indian state
+    - Contains a PAN (alphanumeric sub-string ranging from the 2nd to the 11th character)
+    - Last character is an alphanumeric
+
+    Reference: https://cleartax.in/s/know-your-gstin
+    """
+    # 15 length, first 2 digits, valid pan, checksum
+    if len(field.data) != 15 or int(field.data[:2]) not in short_codes or not field.data[2:12].isalnum() or not field.data[-1].isalnum():
+        raise forms.validators.StopValidation(__("Invalid GSTIN"))
```
jace: "This does not appear to be a valid GSTIN"

boxoffice/forms/order.py
```diff
@@ -68,3 +69,52 @@ class RefundTransactionForm(forms.Form):
...
+
+
+class InvoiceForm(forms.Form):
+    buyer_taxid = forms.StringField(__("GSTIN"), validators=[forms.validators.Optional(),
+        forms.validators.Length(max=255), validate_gstin], filters=[forms.filters.strip(), forms.filters.none_if_empty()])
+    invoicee_name = forms.StringField(__("Full name"), validators=[forms.validators.DataRequired(__("Please enter the buyer's fullname")),
+        forms.validators.Length(max=255)], filters=[forms.filters.strip()])
```
jace: You need the buyer's name and company's name as separate fields for an invoice. Company name should be optional (individuals can get a GST registration).

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,73 @@
...
+__all__ = ['Invoice']
+
+
+def get_latest_invoice_no(organization, jurisdiction, invoice_dt):
+    """
+    Returns the last invoice number used, 0 if no order has been invoiced yet
+    """
+    query = db.session.query(sql.functions.max(Invoice.invoice_no))\
+        .filter(Invoice.organization == organization)
+    fy_start_at, fy_end_at = get_financial_year(jurisdiction, invoice_dt)
```
jace: FY stands for fiscal year, not financial year. The function needs to be renamed. (Although the term financial year is fairly common too.)

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,73 @@
...
+    if fy_start_at and fy_end_at:
```
jace: This `if` condition is unnecessary. The function should _always_ return dates, defaulting to the calendar year.

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,73 @@
...
+        query = query.filter(Invoice.invoiced_at >= fy_start_at, Invoice.invoiced_at < fy_end_at)
+    return query.scalar() or 0
+
+
+class Invoice(UuidMixin, BaseMixin, db.Model):
+    __tablename__ = 'invoice'
+    __uuid_primary_key__ = True
+    __table_args__ = (db.UniqueConstraint('organization_id', 'invoice_no'),)
+
+    invoicee_name = db.Column(db.Unicode(255), nullable=True)
```
jace: Add `invoicee_company`

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,73 @@
...
+    invoicee_email = db.Column(db.Unicode(254), nullable=True)
+    invoice_no = db.Column(db.Integer(), nullable=True)
```
jace: You need to store sequence number and display number (ie, with an alphanumeric prefix) as separate fields, in case the display format changes in future.

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,73 @@
...
+    invoiced_at = db.Column(db.DateTime, nullable=True)
+    street_address_1 = db.Column(db.Unicode(255), nullable=True)
+    street_address_2 = db.Column(db.Unicode(255), nullable=True)
+    city = db.Column(db.Unicode(255), nullable=True)
+    state = db.Column(db.Unicode(255), nullable=True)
+    # India specific: this is the state short code. Eg: KA for Karnataka
+    state_code = db.Column(db.Unicode(4), nullable=True)
```
jace: This is not India specific. We _currently_ only need it for India. Use [ISO 3166-2 codes](https://en.wikipedia.org/wiki/ISO_3166-2), which require _up to_ three digits (not 4).

boxoffice/models/invoice.py
```diff
@@ -0,0 +1,73 @@
...
+        organization = kwargs.get('organization')
+        country_code = kwargs.get('country_code')
+        if not country_code:
+            # Default to India
+            country_code = u'IN'
+        if not organization:
+            raise ValueError(u"Invoice MUST be initialized with an organization")
+        self.invoiced_at = datetime.utcnow()
+        self.invoice_no = get_latest_invoice_no(organization, country_code, self.invoiced_at) + 1
+        super(Invoice, self).__init__(*args, **kwargs)
```
jace: I'm not sure it's a good idea to adopt an invoice number at object construction time. This object may or may not be committed. Look at how [Coaster's BaseScopedIdMixin does it](https://github.com/hasgeek/coaster/blob/master/coaster/sqlalchemy.py#L774) instead.

boxoffice/models/order.py
jace: As above, use the server-side construct from BaseScopedIdMixin to move order number generation into the server.

boxoffice/models/user.py
```diff
@@ -38,3 +60,25 @@ def permissions(self, user, inherited=None):
...
+            start_year = dt.year - 1
+        else:
+            start_year = dt.year
+        # starts on April 1 XXXX
+        fy_start = datetime(start_year, 4, 1)
+        # ends on April 1 XXXX + 1
+        fy_end = datetime(start_year + 1, 4, 1)
+        timezone = u'Asia/Kolkata'
+        return (naive_to_utc(fy_start, timezone), naive_to_utc(fy_end, timezone))
+    return (None, None)
```
jace: Return (Jan 1, Jan 1 next year)

boxoffice/forms/order.py
vidya-ram: Fixed

boxoffice/forms/order.py
vidya-ram: Added `invoicee_company`

boxoffice/models/invoice.py
shreyas-satish: Changed to `get_fiscal_year`.

boxoffice/models/invoice.py
shreyas-satish: Fixed.

boxoffice/models/user.py
shreyas-satish: Fixed.

PR #169, opened 2017-07-12
Fixed travis build error

iambibhas:
Previously `nosetests` had to be told where the tests are explicitly, I removed that when updating travis config with runtests. added `where` config to `setup.cfg` to fix it.

Also removed some redundant `drop_all()` and `create_all()` statements. `init_data()` does both. All the tests call `init_data()` in their `setUp()` function.


tests/test_kharcha.py
```diff
@@ -13,8 +13,6 @@ class TestKharchaAPI(unittest.TestCase):
     def setUp(self):
         self.ctx = app.test_request_context()
         self.ctx.push()
-        db.drop_all()
-        db.create_all()
```
jace: We have to drop and create the db structure between tests to ensure there's no overlap between test cases.

iambibhas: @jace the next line calls `init_data()` which does both `drop_all()` and `create_all()`

iambibhas: All the tests call `init_data()` in their `setUp()` function.

PR #173, opened 2017-09-05
adds validation to prevent finalized invoices from being modified

shreyas-satish:


boxoffice/models/invoice.py
```diff
@@ -72,6 +73,18 @@ def __init__(self, *args, **kwargs):
...
+    @property
+    def is_final(self):
+        return self.status == INVOICE_STATUS.FINAL
+
+@event.listens_for(Invoice, 'before_update')
+def validate_immutable_finalized_invoice(mapper, connection, target):
+    """
+    Raises ValueError if a finalized invoice is being modified. Allows for the status to be toggled.
+    """
+    if target.is_final and get_history(target, 'status').deleted.count(INVOICE_STATUS.DRAFT) == 0:
```
jace: What does `get_history` do? Can you document what's happening here with a comment?

boxoffice/models/invoice.py
```diff
@@ -72,6 +73,20 @@ def __init__(self, *args, **kwargs):
...
+        return self.status == INVOICE_STATUS.FINAL
+
+@event.listens_for(Invoice, 'before_update')
+def validate_immutable_finalized_invoice(mapper, connection, target):
+    """
+    Raises ValueError if a finalized invoice is being modified. Allows the status to be toggled.
+    """
+    # get_history helps check if the invoice was previously in draft status.
+    # This check is necessary to allow edits to an invoice which may be transitioning from draft to final
+    if target.is_final and get_history(target, 'status').deleted.count(INVOICE_STATUS.DRAFT) == 0:
```
jace: Looks like you need `if inspect(target).attrs.status.loaded_value == INVOICE_STATUS.FINAL:`

jace: > If the value has not been loaded, or is otherwise not present in the object’s dictionary, returns NO_VALUE.

So `if inspect(target).attrs.status.loaded_value in (INVOICE_STATUS.FINAL, symbol('NO_VALUE')):`

This will prevent any changes to the `Invoice` if the `status` field was not loaded from the database, which I guess is okay.

boxoffice/models/invoice.py
```diff
@@ -72,6 +73,17 @@ def __init__(self, *args, **kwargs):
...
 
+    @property
+    def is_final(self):
+        return self.status == INVOICE_STATUS.FINAL
+
+    @validates('invoicee_name', 'invoicee_company', 'invoicee_email', 'invoice_no', 'invoiced_at',
+    'street_address_1', 'street_address_2', 'city', 'state', 'state_code', 'country_code', 'postcode',
+    'buyer_taxid', 'seller_taxid', 'customer_order_id', 'organization_id')
+    def validate_immutable_final_invoice(self, key, val):
+        if self.status is INVOICE_STATUS.FINAL:
```
jace: Use `==` when comparing integers. The `is` operator only works for small numbers due to a quirk of CPython optimisation.

boxoffice/views/order.py
```diff
@@ -336,15 +336,21 @@ def edit_invoice_details(order):
     invoice_dict = request.json.get('invoice')
     if not request.json or not invoice_dict:
         return api_error(message=_(u"Missing invoice details"), status_code=400)
+
+    invoice = Invoice.query.get(request.json.get('invoice_id'))
+    if invoice.is_final:
+        return api_error(message=_(u"This invoice has already been generated and hence cannot be modified"),
```
jace: Shouldn't you say "…has been finalised and hence cannot be modified?" The "generated" term is ambiguous.

PR #179, opened 2017-10-16
make title customizable, read discount coupons from parent window if …

shreyas-satish:
Changes:
- Makes the `title` on the ticket listing page optional. This is useful if this page is embedded as an iframe.
- If the ticketing widget is embedded in an iframe, discount coupons are checked for in the parent window.

### DIFF
--- a/boxoffice/views/item_collection.py
+++ b/boxoffice/views/item_collection.py
@@ -79,6 +79,8 @@ def item_collection(item_collection):
     (ItemCollection, {'name': 'item_collection_name', 'organization': 'organization'}, 'item_collection')
     )
 def item_collection_listing(organization, item_collection):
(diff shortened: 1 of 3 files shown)


boxoffice/views/item_collection.py
```diff
@@ -79,6 +79,8 @@ def item_collection(item_collection):
     (ItemCollection, {'name': 'item_collection_name', 'organization': 'organization'}, 'item_collection')
     )
 def item_collection_listing(organization, item_collection):
+    show_title = False if request.args.get('show_title') == 'f' else True
```
jace: Use `show_title = getbool(request.args.get('show_title', True))`

shreyas-satish: Fixed 👍

PR #181, opened 2017-10-26
Better subject lines for order emails

karthikb351:


jace:
What has changed? The two seem identical.

jace:
Oh right, previous was no subject parameter.

karthikb351:
@jace Right all the emails have the same subject line "Thank you for your order!". This includes the item collection and the receipt number in the subject.

PR #184, opened 2017-10-26
Settlement report

shreyas-satish:
Changes:

- Added a `previous_id` to `LineItem` to track versions properly, and a corresponding data migration to update the `previous_id`s retroactively.
- Added a method to retrieve the set of line items that were part of the first payment transaction.
- Added a `pg_refundid` to `PaymentTransaction` to track the refund id issued by the payment gateway.
- Added an integration for Razorpay which enables retrieval of settlements
- Added a method to map the settlements provided by Razorpay, with the transactions in Boxoffice
- Added UI to download the settlement report.


boxoffice/models/line_item.py
```diff
@@ -22,6 +22,8 @@ class LINE_ITEM_STATUS(LabeledEnum):
     #: a line item. Eg: a discount no longer applicable on a line item as a result of a cancellation
     VOID = (3, __("Void"))
 
+LINE_ITEM_STATUS.TRANSACTION = [LINE_ITEM_STATUS.CONFIRMED, LINE_ITEM_STATUS.VOID, LINE_ITEM_STATUS.CANCELLED]
```
jace: This is a kludge. I'm adding a fix to Coaster to support this directly inside the class.

boxoffice/models/line_item.py
```diff
@@ -325,6 +334,11 @@ def get_confirmed_line_items(self):
 Order.get_confirmed_line_items = property(get_confirmed_line_items)
 
 
+def get_initial_line_items(self):
+    return LineItem.query.filter(LineItem.order == self, LineItem.previous == None, LineItem.status.in_(LINE_ITEM_STATUS.TRANSACTION))
+Order.get_initial_line_items = property(get_initial_line_items)
```
jace: If it's a property it shouldn't have the `get_` prefix. It'll be used as `order.initial_line_items.all()`, similar to how a lazy relationship will be.

boxoffice/models/payment.py
```diff
@@ -167,3 +169,72 @@ def calculate_weekly_refunds(item_collection_ids, user_tz, year):
...
+        elif settled_transaction['type'] == 'payment':
+            payment = OnlinePayment.query.filter_by(pg_paymentid=settled_transaction['entity_id']).one_or_none()
+            if payment:
+                order = payment.order
+                rows.append({
+                    'settlement_id': settled_transaction['settlement_id'],
+                    'transaction_type': settled_transaction['type'],
+                    'order_id': order.id,
+                    'payment_id': settled_transaction['entity_id'],
+                    'razorpay_fees': settled_transaction['fee'],
```
jace: PG reference shouldn't be here.

boxoffice/models/payment.py
```diff
@@ -167,3 +169,72 @@ def calculate_weekly_refunds(item_collection_ids, user_tz, year):
         ordered_week_refunds[int(week_refund.sales_week)] = week_refund.sum
 
     return ordered_week_refunds
+
+def get_settled_transactions(date_range):
+    settled_transactions = get_settlements(date_range)
+    headers = ['settlement_id', 'transaction_type', 'order_id', 'payment_id', 'refund_id', 'item_collection', 'description', 'base_amount', 'discounted_amount', 'final_amount', 'order_paid_amount', 'transaction_date', 'settled_at', 'razorpay_fees', 'order_amount', 'credit', 'debit', 'receivable_amount', 'settlement_amount', 'buyer_fullname']
```
jace: Long line needs wrapping.

boxoffice/models/payment.py
```diff
@@ -167,3 +169,72 @@ def calculate_weekly_refunds(item_collection_ids, user_tz, year):
...
+    rows = []
```
jace: Add comment indicating content.

boxoffice/models/payment.py
```diff
@@ -167,3 +169,72 @@ def calculate_weekly_refunds(item_collection_ids, user_tz, year):
...
+    external_transaction_msg = u"Transaction external to Boxoffice. Credited directly to Razorpay?"
```
jace: PG-specific message.

boxoffice/models/payment.py
```diff
@@ -167,3 +169,72 @@ def calculate_weekly_refunds(item_collection_ids, user_tz, year):
...
+            payment = OnlinePayment.query.filter_by(pg_paymentid=settled_transaction['entity_id']).one_or_none()
+            if payment:
+                order = payment.order
+                rows.append({
+                    'settlement_id': settled_transaction['settlement_id'],
+                    'transaction_type': settled_transaction['type'],
+                    'order_id': order.id,
+                    'payment_id': settled_transaction['entity_id'],
+                    'razorpay_fees': settled_transaction['fee'],
+                    'transaction_date': localize_timezone(order.paid_at, 'Asia/Kolkata'),
```
jace: Does Razorpay do IST as a documented feature of their API? Link to docs will help.

migrations/versions/66b67130c901_retroactively_migrate_previous_id_for_.py
```diff
@@ -0,0 +1,109 @@
...
+    nearest_ts = None
+    for ts in lst:
+        diff = abs(timestamp - ts).total_seconds()
+        if diff < 1:
+            if nearest_ts and diff > abs(timestamp - nearest_ts).total_seconds():
+                # there's a timestamp that's closer
+                pass
+            else:
+                nearest_ts = ts
+    return nearest_ts
```
jace: This could be simplified to ([source](https://stackoverflow.com/a/12141207/78903)):

```python
nearest_ts = min(lst, key=lambda x: abs(x - timestamp).total_seconds())
if abs(nearest_ts - timestamp).total_seconds() < 1:
    return nearest_ts
```

migrations/versions/66b67130c901_retroactively_migrate_previous_id_for_.py
```diff
@@ -0,0 +1,109 @@
...
+
+def set_previous_keys_for_line_items(line_items):
+    timestamped_line_items = OrderedDict()
+    
+    # Assemble the `timestamped_line_items` dictionary with the timestamp at which the line items were created
+    # as the key, and the line items that were created at that time as the value (as a list)
+    # Some line items may have been created a few milliseconds later, so the nearest timestamp
+    # with a tolerance level of one second is searched for
```
jace: This depends on whether we used `datetime.utcnow()` or `func.utcnow()`. The latter won't have this problem as it uses the transaction's timestamp. Maybe add a comment explaining that this may be caused by having used `datetime.utcnow()` in the past.

migrations/versions/66b67130c901_retroactively_migrate_previous_id_for_.py
```diff
@@ -0,0 +1,109 @@
...
+            'item_id': line_item.item_id,
+            'previous_id': None
+        })
+
+    # The previous line item for a line item, is a line item that has an earlier timestamp with a void status
+    # with the same item_id. Find it and set it
+    used_line_item_ids = set()
+    for idx, (timestamp, line_item_dicts) in enumerate(timestamped_line_items.items()):
+        # 0th timestamps are root line items
+        if idx > 0:
```
jace: Why not just use `enumerate(timestamped_line_items.items())[1:]` and skip the `if`?

jace: Also, I'm unclear on what `timestamped_line_items` contains and why the first element can be ignored.

migrations/versions/66b67130c901_retroactively_migrate_previous_id_for_.py
```diff
@@ -0,0 +1,109 @@
...
+            for li_dict in timestamped_line_items[timestamp]:
+                previous_li_dict = [previous_li_dict
+                    for previous_li_dict in timestamped_line_items[timestamped_line_items.keys()[idx-1]]
+                    if previous_li_dict['item_id'] == li_dict['item_id']
+                    and previous_li_dict['id'] not in used_line_item_ids
+                    and previous_li_dict['status'] == LINE_ITEM_STATUS.VOID][0]
```
jace: Indent the two `and` lines and move the closing `][0]` to its own line. This took a while to parse.

migrations/versions/66b67130c901_retroactively_migrate_previous_id_for_.py
shreyas-satish: The 0th group of `line_item_dicts` is the group of line items that comprised of the initial transaction. Hence their `previous_id` will remain `None`. It's only the subsequent groups that will need their `previous_id` updated.

boxoffice/models/line_item.py
```diff
@@ -325,6 +332,11 @@ def get_confirmed_line_items(self):
 Order.get_confirmed_line_items = property(get_confirmed_line_items)
 
 
+def initial_line_items(self):
+    return LineItem.query.filter(LineItem.order == self, LineItem.previous == None, LineItem.status.in_(LINE_ITEM_STATUS.TRANSACTION))
+Order.initial_line_items = property(initial_line_items)
```
jace: Both this and the other function could be lazy relationships with elaborate `primaryjoin` conditions. Seems cleaner.

boxoffice/models/payment.py
```diff
@@ -167,3 +169,5 @@ def calculate_weekly_refunds(item_collection_ids, user_tz, year):
         ordered_week_refunds[int(week_refund.sales_week)] = week_refund.sum
 
     return ordered_week_refunds
+
+
```
jace: Extra blank lines

boxoffice/models/user.py
```diff
@@ -54,6 +54,9 @@ class Organization(ProfileBase, db.Model):
     # logo (image url), refund_policy (html), ticket_faq (html), website (url)
     details = db.Column(JsonDict, nullable=False, server_default='{}')
     contact_email = db.Column(db.Unicode(254), nullable=False)
+    # Merchant account involves settlements and hence can access
+    # settlement reports
+    merchant = db.Column(db.Boolean, nullable=False, default=False)
```
jace: What sort of organization accounts would we have that are not merchant accounts?

boxoffice/views/admin_report.py
```diff
@@ -169,3 +170,23 @@ def row_handler(row):
...
+        abort(401)
+    year = int(request.args.get('year'))
+    month = int(request.args.get('month'))
+    try:
+        date(year, month, 1)
+    except (ValueError, TypeError):
+        return api_error(message='Invalid year/month',
+                status_code=403,
+                errors=['invalid_date'])
+    headers, rows = get_settled_transactions({'year': year, 'month': month}, g.user.timezone)
```
jace: This call does not specify the organization it's being called for. Since the RazorPay API key is global to the app, there is no reason to drag the organization into this. Just use Lastuser's `siteadmin` permission with a `@lastuser.requires_permission('siteadmin')` on this view. Or use a new permission name if `siteadmin` is too much to give away.

PR #198, opened 2017-12-15
Allows hierarchy in organization

shreyas-satish:
Adds a `parent_id` attribute to `Organization` to allow organizations to defer invoicing to their parent organization.


jace:
`parent_id` will clash with hasgeek/lastuser#6. Can you make this `invoicer_id`?

PR #200, opened 2018-01-08
Add functionality to raise a custom event once boxoffice.js has been loaded.

vidya-ram:
Add functionality to raise a custom event once boxoffice.js has been loaded

On the client website where Boxoffice ticketing widget needs to be embedded add event handler for the custom event `boxofficeScript-load`

  `window.addEventListener('boxofficeScript-load', function (e) {
    window.Boxoffice.init({
      org: "{{ org_title }}",
      itemCollection: "{{ boxoffice_item_collection }}",
      paymentDesc: "{{ item_collection_title }}"
    });`
  }, false);

### DIFF
--- a/boxoffice/templates/boxoffice.js
+++ b/boxoffice/templates/boxoffice.js
@@ -285,8 +285,7 @@ $(function() {
           if (response.readyState === 4) {
           	//Server error
             onServerError();
-          }
-          else if (response.readyState === 0) {
+          } else if (response.readyState === 0) {
             if (ajaxLoad.retries < 0) {
               //Network error
               onNetworkError();
@@ -306,8 +305,7 @@ $(function() {
                 formDetails[this.name] = [formDetails[this.name]];
               }
               formDetails[this.name].push(this.value || '');
-            }
-            else {
+            } else {
               formDetails[this.name] = this.value || '';
             }
           });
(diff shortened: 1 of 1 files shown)


boxoffice/templates/boxoffice.js
```diff
@@ -888,5 +888,17 @@ $(function() {
   window.Boxoffice.init = function(widgetConfig) {
 	  boxoffice.init(widgetConfig);
   };
+
+  // Raise a custom event once boxoffice.js has been loaded
+  if (typeof(window.Event) === "function") {
+    var event = new Event('boxofficeScript-load');
```
shreyas-satish: Can we use `on*` as a convention when triggering events. So, that would be `onBoxofficeInit` as the event name and `onBoxofficeInit` as the variable name.

boxoffice/templates/boxoffice.js
```diff
@@ -888,5 +888,17 @@ $(function() {
...
+  }
```
shreyas-satish: Can you please collapse these two lines into a single line, since we're using the following convention:
```
if (condition) {
 //
} else {
  //
}
```

boxoffice/templates/boxoffice.js
vidya-ram: Fixed

boxoffice/templates/boxoffice.js
vidya-ram: Fixed

PR #202, opened 2018-01-19
Add support to organization admin page to create a new item collection

vidya-ram:
This PR is depends on Baseframe PR https://github.com/hasgeek/baseframe/pull/176


boxoffice/views/admin.py
```diff
@@ -24,11 +25,27 @@ def index():
...
+    if ic:
+        return {
+            'id': ic.id,
+            'name': ic.name,
+            'title': ic.title,
+            'url': '/ic/' + unicode(ic.id),
+            'description_text': ic.description_text,
+            'description_html': ic.description_html
+        }
+    return None
```
shreyas-satish: This line isn't required. `None` is returned by default.

boxoffice/views/admin.py
```diff
@@ -24,11 +25,27 @@ def index():
     return dict(user=g.user)
 
 
+def jsonify_ic(ic):
+    if ic:
+        return {
+            'id': ic.id,
+            'name': ic.name,
+            'title': ic.title,
+            'url': '/ic/' + unicode(ic.id),
```
shreyas-satish: Better to use `url_for`.

boxoffice/assets/js/views/org.js
```diff
@@ -2,30 +2,89 @@
...
+      btnDisable: true,
+      btnEnable: false,
+      empty: ""
+    };
+
     fetch({
       url: urlFor('view', {resource: 'o', id: org_name, root: true})
-    }).then(function({id, org_title, item_collections}){
+    }).then(function({id, org_title, item_collections, form}) {
+      let AddICFormComponent = Ractive.extend({
```
shreyas-satish: Seems like too long a name, `ICForm` should suffice.

boxoffice/assets/js/views/org.js
```diff
@@ -2,30 +2,89 @@
...
+        isolated: false,
+        template: Util.getFormTemplate(form),
+        data: {
+          addFormId: Util.getElementId(form),
```
shreyas-satish: `addFormId` seems like a name for a function, but here it's just a variable. `formId` perhaps?

boxoffice/assets/js/models/util.js
```diff
@@ -28,6 +28,17 @@ export const Util = {
     else {
       return moment(dateTimeString).toString();
     }
+  },
+  getElementId: function(htmlString) {
+    return htmlString.match(/id="(.*?)"/)[1];
+  },
+  getFormTemplate: function(htmlString) {
+    // Add on click event handler for Ractive to submit the form
+    const formTemplate = `${htmlString.slice(0, htmlString.search(/type="submit"/))} on-click="onFormSubmit(event)" {{#formOnSubmit}}disabled{{/}} ${htmlString.slice(htmlString.search(/type="submit"/))}`;
```
shreyas-satish: Can you break this into multiple lines since it's too long? Also you can return the value directly, `formTemplate` isn't really needed.

boxoffice/assets/js/models/util.js
```diff
@@ -28,6 +28,17 @@ export const Util = {
...
+    return formTemplate
+  },
+  showFormErrors: function(formId, errors) {
```
shreyas-satish: Is this function required? Can call the underlying function directly, since this function just passes the call through?

boxoffice/assets/js/models/util.js
```diff
@@ -43,6 +54,7 @@ export const post = function (config) {
     url: config.url,
     type: 'POST',
     data: config.data,
+    processData: config.processData ? config.processData : false,
```
shreyas-satish: Can you add a comment to describe what this means?

boxoffice/views/admin.py
vidya-ram: Fixed

boxoffice/views/admin.py
vidya-ram: Fixed

boxoffice/assets/js/views/org.js
vidya-ram: Fixed

boxoffice/assets/js/views/org.js
vidya-ram: Fixed

boxoffice/assets/js/models/util.js
vidya-ram: Fixed

boxoffice/assets/js/models/util.js
vidya-ram: Fixed

boxoffice/forms/item_collection.py
```diff
@@ -0,0 +1,14 @@
...
+
+__all__ = ['ItemCollectionForm']
+
+
+class ItemCollectionForm(forms.Form):
+    title = forms.StringField(__("Item Collection title"),
+        validators=[forms.validators.DataRequired(__("Please specify a title")),
+        forms.validators.Length(max=250)], filters=[forms.filters.strip()])
+    description = forms.TinyMce4Field(__("Description"))
+    tax_type = forms.SelectField(u"Tax type", choices=[('GST', 'GST')], default="GST")
```
shreyas-satish: Both choices are `GST`?

vidya-ram: It is just one option. `choices` needs to be tuple (value, label)

boxoffice/assets/js/models/util.js
vidya-ram: Fixed

boxoffice/assets/js/views/org.js
```diff
@@ -2,30 +2,88 @@
...
+    }).then(function({id, org_title, item_collections, form}) {
+      let ICForm = Ractive.extend({
+        isolated: false,
+        template: Util.getFormTemplate(form),
+        data: {
+          formId: Util.getElementId(form),
+        },
+        onFormSubmit: function(event) {
+          event.original.preventDefault();
+          let self = this;
```
shreyas-satish: Any reason why `self` is required here? This shouldn't be required if you use the arrow function instead?

boxoffice/assets/js/models/util.js
```diff
@@ -28,6 +28,14 @@ export const Util = {
...
       return moment(dateTimeString).toString();
     }
+  },
+  getElementId: function(htmlString) {
+    return htmlString.match(/id="(.*?)"/)[1];
+  },
+  getFormTemplate: function(htmlString) {
+    // Add on click event handler for Ractive to submit the form
+    return `${htmlString.slice(0, htmlString.search(/type="submit"/))} on-click="onFormSubmit(event)" 
+      {{#formOnSubmit}}disabled{{/}} ${htmlString.slice(htmlString.search(/type="submit"/))}`;
```
shreyas-satish: Where is `formOnSubmit` coming from?

vidya-ram: Do you mean it is better to pass the submit event handler as an argument to `getFormTemplate`?

shreyas-satish: Yes. A new developer shouldn't have to understand the internals of this function to use it. The inputs, outputs and side effects (if any) should be obvious.

boxoffice/assets/js/views/org.js
vidya-ram: Fixed

boxoffice/assets/js/models/util.js
vidya-ram: Fixed

boxoffice/__init__.py
```diff
@@ -43,6 +43,7 @@
 lastuser.init_usermanager(UserManager(db, User))
 app.config['tz'] = timezone(app.config['TIMEZONE'])
 baseframe.init_app(app, requires=['boxoffice'], ext_requires=['baseframe-bs3', 'fontawesome>=4.0.0', 'baseframe-footable'])
+app.assets.register('js_tinymce', assets.require('!jquery.js', 'tinymce.js>=4.0.0', 'jquery.tinymce.js>=4.0.0'))
```
jace: This registration should be in Baseframe since it's a shared asset.

boxoffice/forms/item_collection.py
jace: `tax_type = forms.SelectField(__("Tax type"), choices=[('GST', 'GST')], default='GST')`

boxoffice/views/admin.py
```diff
@@ -24,11 +25,26 @@ def index():
     return dict(user=g.user)
 
 
+def jsonify_ic(ic):
```
jace: Since this returns a dictionary and not JSON, the name should be `as_dict`. Also, this should be a method on the ItemCollection, not an independent function.

boxoffice/views/admin.py
```diff
@@ -24,11 +25,26 @@ def index():
...
+    if ic:
```
jace: This condition is unnecessary since`ic` is a required parameter.

boxoffice/views/admin.py
```diff
@@ -24,11 +25,26 @@ def index():
...
+        return {
+            'id': ic.id,
+            'name': ic.name,
+            'title': ic.title,
+            'url': url_for('admin_item_collection', ic_id=ic.id),
```
jace: Use `self.url_for('admin_item_collection')`. If the `url_for` method doesn't recognise that action, add it.

boxoffice/views/admin.py
jace: Actually, you don't even need this. RoleMixin's `access_for` should construct the dictionary for you. @shreyas-satish can you help with this?

boxoffice/views/admin.py
```diff
@@ -24,11 +25,26 @@ def index():
...
+
+
 def jsonify_org(data):
     item_collections_list = ItemCollection.query.filter(ItemCollection.organization == data['org']).order_by('created_at desc').all()
+    form = ItemCollectionForm()
+    html_form = render_form(form=form, title=u"New Item Collection", submit=u"Save", ajax=False, with_chrome=False)
     return jsonify(id=data['org'].id,
         org_title=data['org'].title,
-        item_collections=[{'id': ic.id, 'name': ic.name, 'title': ic.title, 'url': '/ic/' + unicode(ic.id), 'description_text': ic.description_text, 'description_html': ic.description_html} for ic in item_collections_list])
+        item_collections=[jsonify_ic(ic) for ic in item_collections_list],
```
jace: `item_collections=[ic.access_for(…) for ic in item_collections_list]` @shreyas-satish

boxoffice/views/admin.py
vidya-ram: We need relative path(without `/admin`) here since backbone is handling the routing

boxoffice/forms/item_collection.py
vidya-ram: Fixed

boxoffice/views/admin.py
shreyas-satish:
```
class ItemCollection(BaseScopedNameMixin, db.Model):
    # ... excluded table details
    __roles__ = {
        'ic_owner': {
            'write': {},
            'read': {'id', 'name', 'title'}
        }
    }

    def roles_for(self, user=None, token=None):
        if not user and not token:
            return set()
        roles = super(ItemCollection, self).roles_for(user, token)
        if user or token:
            roles.add('user')
        if self.organization.userid in user.organizations_owned_ids():
            roles.add('ic_owner')
        return roles

Example usage:
=> ic = ItemCollection.query.first()
=> dict(ic.access_for(user=g.user))
```

boxoffice/__init__.py
vidya-ram: Fixed

boxoffice/views/admin.py
vidya-ram: Fixed

boxoffice/views/admin.py
vidya-ram: Fixed

boxoffice/assets/js/views/org.js
```diff
@@ -2,33 +2,79 @@
...
+            url: urlFor('new', {
+              scope_ns: 'o',
+              scope_id: org_name,
+              resource: 'ic',
+              root: true
+            }),
+            data: getFormParameters(formSelector),
+            formId: formSelector
+          }).done((remoteData) => {
+            orgComponent.unshift('item_collections', remoteData.result.item_collection);
```
shreyas-satish: Are these two lines required since we're navigating to the dashboard directly?

boxoffice/models/item_collection.py
```diff
@@ -19,3 +20,22 @@ class ItemCollection(BaseScopedNameMixin, db.Model):
     organization = db.relationship(Organization, backref=db.backref('item_collections', cascade='all, delete-orphan'))
     parent = db.synonym('organization')
     tax_type = db.Column(db.Unicode(80), nullable=True, default=u'GST')
+
+    @property
```
shreyas-satish: This is better off as a function, since that's the convention we're using. Also, this function will take a parameter. Let's keep this definition consistent with how we're using it elsewhere?

boxoffice/views/admin.py
```diff
@@ -26,9 +27,12 @@ def index():
 
 def jsonify_org(data):
     item_collections_list = ItemCollection.query.filter(ItemCollection.organization == data['org']).order_by('created_at desc').all()
+    form = ItemCollectionForm()
```
shreyas-satish: This variable isn't required since it's only being used once. Invoke `ItemCollectionForm()` directly on L31?

vidya-ram: Fixed

boxoffice/assets/js/views/org.js
vidya-ram: Fixed

boxoffice/models/item_collection.py
vidya-ram: Fixed

boxoffice/views/admin.py
```diff
@@ -26,9 +27,11 @@ def index():
 
 def jsonify_org(data):
     item_collections_list = ItemCollection.query.filter(ItemCollection.organization == data['org']).order_by('created_at desc').all()
+    html_form = render_form(form=ItemCollectionForm(), title=u"New Item Collection", submit=u"Save", ajax=False, with_chrome=False)
```
shreyas-satish: `ajax` is `False` by default in `render_form` so it isn't required as an argument.

boxoffice/assets/js/models/util.js
```diff
@@ -97,6 +105,29 @@ export const getCsrfToken = function () {
   return document.head.querySelector("[name=csrf-token]").content;
 };
 
+export const formErrorHandler = function(errorResponse, formId) {
+  let errorMsg = "";
+  if (errorResponse.readyState === 4) {
```
shreyas-satish: Can you add a comment that explains what `4` here means?

boxoffice/views/admin.py
```diff
@@ -42,6 +45,26 @@ def org(organization):
...
+    ic_form = ItemCollectionForm()
+    if ic_form.validate_on_submit():
+        ic = ItemCollection(organization=organization)
+        ic_form.populate_obj(ic)
+        if not ic.name:
+            ic.make_name()
+        db.session.add(ic)
+        db.session.commit()
+        return api_success(result={'item_collection': dict(ic.access_for(user=g.user))}, doc=_(u"New item collection created"), status_code=201)
+    return api_error(message=_(u"Incorrect data"), errors=ic_form.errors, status_code=400)
```
shreyas-satish: `Incorrect data` is a little vague. Better to return the form's validation messages?

PR #207, opened 2018-02-05
make compatible with coaster#161

shreyas-satish:
Replaced `user` with `actor` and `token` with `anchors` in `roles_for` as per https://github.com/hasgeek/coaster/pull/161/files.

### DIFF
--- a/boxoffice/models/item_collection.py
+++ b/boxoffice/models/item_collection.py
@@ -37,8 +37,6 @@ class ItemCollection(BaseScopedNameMixin, db.Model):
     }
 
     def roles_for(self, actor=None, anchors=()):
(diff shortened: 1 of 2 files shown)


boxoffice/models/item_collection.py
```diff
@@ -36,10 +36,10 @@ def url_for_view(self):
         }
     }
 
-    def roles_for(self, user=None, token=None):
-        if not user and not token:
+    def roles_for(self, actor=None, anchors=()):
+        if not actor and not anchors:
             return set()
```
jace: Remove this. Always call `super`, since you need the `all` and `auth` roles acquired from there.

PR #211, opened 2018-02-22
Form view

shreyas-satish:
This PR:

- Introduces `FormView`, a component to help create and edit a specific resource's details in a slider (modal support to come later).
- Uses `FormView` to help create and edit an item collection.
- Introduces `data-navigate`, which is intended to be used to decorate an anchor tag in order to trigger a page transition with `pushState`.
- Initiates a move towards co-locating the JavaScript component and the corresponding template.


boxoffice/assets/js/templates/item_collection.html.js
```diff
@@ -59,6 +59,11 @@ export const AggChartTemplate = `
 export const ItemCollectionTemplate = `
   <div class="content-wrapper">
     <h1 class="header">{{ icTitle }}</h1>
+    <div class="title-wrapper col-xs-12">
+      <a class="boxoffice-button boxoffice-button-action btn-right" href="/admin/ic/{{ic_id}}/edit" data-navigate>
+        Edit item collection
```
jace: How do we do localization in JavaScript-rendered templates?

boxoffice/views/order.py
```diff
@@ -402,6 +402,7 @@ def format_assignee(line_item):
             'phone': line_item.current_assignee.phone
         }
 
+        # if line_item.item.assignee_details:
```
jace: ?

boxoffice/views/admin_item_collection.py
```diff
@@ -59,3 +62,54 @@ def admin_item_collection(item_collection):
...
+def admin_new_ic(organization):
+    return dict(organization=organization)
+
+
+def jsonify_edit_item_collection(item_collection_dict):
+    item_collection = item_collection_dict['item_collection']
+    ic_form = ItemCollectionForm(obj=item_collection)
+    if request.method == 'GET':
+        return jsonify(form_template=render_form(form=ic_form,
+            title=u"Edit item collection", submit=u"Save", ajax=False, with_chrome=False))
```
jace: This may be a good time to expose Baseframe's [FormGenerator](https://github.com/hasgeek/baseframe/blob/6c94f5b07eb279519a0d8ffe3c183df42ab1e82c/baseframe/forms/form.py#L128-L132) as an API. (Separate ticket)

boxoffice/assets/js/views/router.js
```diff
@@ -36,6 +40,14 @@ export const Router = Backbone.Router.extend({
   item_collection: function(ic_id) {
     ItemCollectionView.render({ic_id});
   },
+  new_item_collection: function(org_name){
+    OrgView.render({org_name});
```
vidya-ram: Is `OrgView` component required here since on successful creation of an item collection, org page is fetched again and loaded.
Also there is a delay in the New item collection (/admin/o/hasgeek/ic/new) page,  the `OrgView` is loaded before before `ItemCollectionNewView` which can lead to confusion for the user

boxoffice/assets/js/views/detail_view.js
```diff
@@ -0,0 +1,104 @@
...
+  handleTemplate: function(response, options){
+    this.set('handleForm', false);
+    this.set('handleTemplate', true);
+    this.set('template', options.template);
+  }
+});
+
+DetailView.on('hide', function(event){
+  this.hide();
+  navigateBack();
```
vidya-ram: window.history.go(-1) will break in the case there is no back history. For eg:- `new` or `edit` page has been loaded directly.

boxoffice/assets/js/views/router.js
```diff
@@ -36,6 +40,14 @@ export const Router = Backbone.Router.extend({
...
+    ItemCollectionNewView.render({org_name});
+  },
+  edit_item_collection: function(ic_id){
+    ItemCollectionView.render({ic_id});
```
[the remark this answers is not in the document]
vidya-ram: This is also leading to unnecessary fetch of item collection details again on clicking `edit item collection` button

boxoffice/assets/js/views/detail_view.js
```diff
@@ -0,0 +1,104 @@
...
+
+
+let DetailViewSliderTemplate = `
+  {{#if shown}}
+    <div class="content-slider align-down" intro-outro='fly:{x:200,y:0,duration:200}'>
+      <button class="close-button" on-click="hide"><i class="fa fa-close"></i></button>
+      <p class="content-slider-title">{{{title}}}</p>
+      <div class="content-slider-wrapper">
+        {{#if handleForm}}
+          {{{ formHTML }}}
```
vidya-ram: JS (inside script tags) sent along with form html from Baseframe will not work in this. They have to passed as templates to Ractive instance.

boxoffice/assets/js/views/detail_view.js
```diff
@@ -0,0 +1,104 @@
...
+    handleForm: false,
+    template: '',
+    // Use this to set custom attributes in the template
+    attributes: {}
+  },
+  load: function(options){
+    fetch({url: options.url}).then((response) => {
+      this.hide();
+      this.set('title', options.title);
+      this.show();
```
vidya-ram: Why is both this.hide() & this.show(); called here?
Also wouldn't it better to set show form to true once the fetch request for form is successful

boxoffice/assets/js/views/edit_item_collection.js
```diff
@@ -0,0 +1,17 @@
...
+import {navigateTo} from '../views/main_admin.js';
+
+export const ItemCollectionEditView = {
+  render: function({ic_id}={}) {
+    DetailView.load({
+      url: urlFor('edit', {resource: 'ic', id: ic_id, root: true}),
+      title: 'Edit item collection',
+      handleForm: true,
+      onSuccess: function(responseData){
+        navigateTo(urlFor('view', {resource: 'ic', id: ic_id, root: true}))
```
vidya-ram: Could we also abstract `onSuccess` out and add it to DetailView

boxoffice/assets/js/views/main_admin.js
```diff
@@ -1,13 +1,42 @@
...
+  return false;
+}
+
+function handleNavigation(){
+  /*
+    To trigger page transitions through pushState, add `data-navigate` to
+    the anchor tag and specify the URL in the `href` attribute
+  */
+  document.addEventListener('click', function(event){
+    event.preventDefault();
```
vidya-ram: Won't this intercept all clicks even those that aren't for navigation ?

boxoffice/assets/js/views/detail_view.js
```diff
@@ -0,0 +1,104 @@
...
+    });
+  },
+  show: function(){
+    this.set('shown', true);
+  },
+  hide: function(){
+    this.set('shown', false);
+  },
+  handleForm: function(response, options){
+    this.set('handleTemplate', false);
```
vidya-ram: Isn't `handleTemplate` a method, why is it set to false here?

boxoffice/assets/js/views/edit_item_collection.js
shreyas-satish: Not necessarily, since `ItemCollectionEditView` (in this case) should decide what needs to happen after a form submit was successful eg: redirect to a particular route.

PR #212, opened 2018-02-28
Item admin

shreyas-satish:
- Added an area graph to visualize the demand curve
- Presents item details and associated details of prices and discount policies


shreyas-satish:
Also added:
- Ability to create and edit items
- Ability to create and edit prices

shreyas-satish:
More changes:
- Added a form to create categories
- Moved methods from `LineItem` to the models on which they were defined on. Moved out `LineItemDiscounter` to a separate file.

boxoffice/models/item.py
```diff
@@ -79,12 +92,79 @@ def quantity_available(self):
...
+        from ..models import LineItem, LINE_ITEM_STATUS
+
+        return LineItem.query.filter(LineItem.item == self, LineItem.status == LINE_ITEM_STATUS.CANCELLED).count()
+
+    @property
+    def net_sales(self):
+        from ..models import LineItem, LINE_ITEM_STATUS
+
+        return db.session.query(func.sum(LineItem.final_amount)).filter(LineItem.item == self,
+            LineItem.status == LINE_ITEM_STATUS.CONFIRMED).first()[0]
```
iambibhas: Doesn't `first()` return a single instance and not a list?

boxoffice/models/discount_policy.py
```diff
@@ -109,6 +110,63 @@ def make_bulk(cls, discount_code_base, **kwargs):
...
+                for coupon in coupons:
+                    if coupon.usage_limit > coupon.used_count:
+                        policies.append((coupon.discount_policy, coupon))
+        return policies
+
+    @property
+    def line_items_count(self):
+        from ..models import LineItem, LINE_ITEM_STATUS
+
+        return LineItem.query.filter(LineItem.discount_policy == self, LineItem.status == LINE_ITEM_STATUS.CONFIRMED).count()
```
iambibhas: Probably for a future PR, but if we added `line_items` as a relationship in this model, instead of the other way around, these queries will get much simpler. ` return self.line_items.filter(LineItem.status == LINE_ITEM_STATUS.CONFIRMED).count()`.

boxoffice/models/item.py
iambibhas: Also, same as discount policy. If `Item` had `line_items` as relationship, these 5 methods above will become much simpler. But this works for now.

shreyas-satish: > Doesn't first() return a single instance and not a list?

No, since the query is on `db.session` and not on a specific model, it returns a tuple.

boxoffice/forms/item.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+
+from baseframe import __
+import baseframe.forms as forms
+from baseframe.forms.sqlalchemy import QuerySelectField
+from ..models import db, Item, Category, ItemCollection
+from sqlalchemy.sql import select, func
```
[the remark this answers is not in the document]
shreyas-satish: Thanks, I've fixed this.

boxoffice/models/item.py
shreyas-satish: Good point, I've added the lazy loader on the relationship and the queries are cleaner as a result.

boxoffice/models/discount_policy.py
shreyas-satish: Fixed.

boxoffice/forms/category.py
```diff
@@ -0,0 +1,15 @@
...
+from baseframe import __
+import baseframe.forms as forms
+
+__all__ = ['CategoryForm']
+
+
+class CategoryForm(forms.Form):
+    title = forms.StringField(__("Category title"),
+        validators=[forms.validators.DataRequired(__("Please specify a title")),
+        forms.validators.Length(max=250)], filters=[forms.filters.strip()])
```
jace: Indent.

boxoffice/forms/item.py
```diff
@@ -0,0 +1,29 @@
...
+from baseframe.forms.sqlalchemy import QuerySelectField
+from ..models import db, Category, ItemCollection
+
+__all__ = ['ItemForm']
+
+
+class ItemForm(forms.Form):
+    title = forms.StringField(__("Item title"),
+        validators=[forms.validators.DataRequired(__("Please specify a title")),
+        forms.validators.Length(max=250)], filters=[forms.filters.strip()])
```
jace: Indent.

boxoffice/forms/price.py
```diff
@@ -0,0 +1,17 @@
...
+
+from baseframe import __
+import baseframe.forms as forms
+
+__all__ = ['PriceForm']
+
+
+class PriceForm(forms.Form):
+    start_at = forms.DateTimeField(__("Start date & time"),
+        validators=[forms.validators.DataRequired(__("Please specify an start date & time"))])
```
jace: "a start"

boxoffice/models/category.py
```diff
@@ -15,5 +15,18 @@ class Category(BaseScopedNameMixin, db.Model):
     seq = db.Column(db.Integer, nullable=False)
     item_collection = db.relationship(ItemCollection,
         backref=db.backref('categories', cascade='all, delete-orphan',
-            order_by=seq, collection_class=ordering_list('seq', count_from=1)))
+        order_by=seq, collection_class=ordering_list('seq', count_from=1)))
```
jace: Indent required since these are parameters to the backref, not to the relationship.

boxoffice/models/discount_policy.py
```diff
@@ -137,6 +193,11 @@ def __init__(self, *args, **kwargs):
     discount_policy_id = db.Column(None, db.ForeignKey('discount_policy.id'), nullable=False)
     discount_policy = db.relationship(DiscountPolicy, backref=db.backref('discount_coupons', cascade='all, delete-orphan'))
 
+    def update_used_count(self):
+        from ..models import LineItem, LINE_ITEM_STATUS
+
+        self.used_count = db.select([db.func.count()]).where(LineItem.discount_coupon == self).where(LineItem.status == LINE_ITEM_STATUS.CONFIRMED).as_scalar()
```
jace: Add a `cached` annotation to `used_count`

boxoffice/models/discount_policy.py
```diff
@@ -137,6 +193,11 @@ def __init__(self, *args, **kwargs):
     discount_policy_id = db.Column(None, db.ForeignKey('discount_policy.id'), nullable=False)
     discount_policy = db.relationship(DiscountPolicy, backref=db.backref('discount_coupons', cascade='all, delete-orphan'))
 
+    def update_used_count(self):
+        from ..models import LineItem, LINE_ITEM_STATUS
```
jace: This import can be at the top of the file. It should work as expected. It will also work if written as `from . import LineItem, LINE_ITEM_STATUS`.

boxoffice/models/item.py
```diff
@@ -44,6 +45,20 @@ class Item(BaseScopedNameMixin, db.Model):
 
     cancellable_until = db.Column(db.DateTime, nullable=True)
 
+    __roles__ = {
+        'item_owner': {
+            'write': {},
```
jace: Surely the item owner can write all these fields? If the `write` action is _unused_ at the moment, leave this out. Less ambiguity that way. RoleMixin will add a blank for you. (Also applies to other places `__roles__` is defined.)

boxoffice/models/item.py
```diff
@@ -79,12 +91,79 @@ def quantity_available(self):
...
-        """Checks if an item has a current price object and has a positive quantity_available"""
+        """Check if an item has a current price object and has a positive quantity_available"""
         return bool(self.current_price() and self.quantity_available > 0)
 
     def is_cancellable(self):
         return datetime.utcnow() < self.cancellable_until if self.cancellable_until else True
 
+    def url_for(self, action='view', _external=False, **kwargs):
+        if action == 'view':
+            return url_for('admin_item', item_id=self.id, _external=_external, **kwargs)
```
jace: I'll recommend using the `@Item.is_url_for('view')` decorator on the view and removing this method entirely.

boxoffice/models/item.py
```diff
@@ -79,12 +91,79 @@ def quantity_available(self):
...
+
+    @property
+    def url_for_view(self):
+        return self.url_for('view')
```
jace: Is this necessary? Using `item.url_for()` should suffice in templates, since 'view' is the default action.

boxoffice/models/item.py
```diff
@@ -79,12 +91,79 @@ def quantity_available(self):
...
+
+    @property
+    def active_price(self):
+        return self.current_price().amount if self.current_price() else None
```
jace: If `current_price()` has a lookup cost, this is calling it twice. Assign to a local variable?

boxoffice/models/item.py
```diff
@@ -79,12 +91,79 @@ def quantity_available(self):
...
+
+    @property
+    def get_confirmed_line_items(self):
```
jace: A property shouldn't have `get_`, since that implies a callable. This can be just `confirmed_line_items`.

boxoffice/models/item.py
```diff
@@ -79,12 +91,79 @@ def quantity_available(self):
...
+        from ..models import LineItem, LINE_ITEM_STATUS
+
+        return self.line_items.filter(LineItem.status == LINE_ITEM_STATUS.CANCELLED).count()
+
+    @property
+    def net_sales(self):
+        from ..models import LineItem, LINE_ITEM_STATUS
+
+        return db.session.query(db.func.sum(LineItem.final_amount)).filter(LineItem.item == self,
+            LineItem.status == LINE_ITEM_STATUS.CONFIRMED).first()[0]
```
jace: These four should not be properties since they have a non-trivial calling cost.

boxoffice/models/item.py
```diff
@@ -105,3 +184,38 @@ class Price(BaseScopedNameMixin, db.Model):
...
+    @property
+    def json_start_at(self):
+        return localize_timezone(self.start_at).isoformat()
+
+    @property
+    def json_end_at(self):
+        return localize_timezone(self.end_at).isoformat()
+
+    @property
+    def tense(self):
```
jace: Not a property since the result changes with time even if the data does not change.

boxoffice/models/order.py
```diff
@@ -114,6 +114,18 @@ def is_fully_assigned(self):
                 return False
         return True
 
+    @property
+    def get_confirmed_line_items(self):
+        from ..models import LineItem, LINE_ITEM_STATUS
+
+        return LineItem.query.filter(LineItem.order == self, LineItem.status == LINE_ITEM_STATUS.CONFIRMED).all()
```
jace: Not a property since it hits the database with `.all()`

shreyas-satish:
Thank you @jace. I've made the suggested changes. In addition, I've removed create and edit functionality for item collections, items, categories and prices from `siteadmin` since it's no longer necessary.

PR #215, opened 2018-03-26
UI for partial refunds

shreyas-satish:
- Added ability to issue partial refunds from the dashboard


boxoffice/forms/order_refund.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+
+from baseframe import __
+import baseframe.forms as forms
+from baseframe.forms.validators import StopValidation
+
+__all__ = ['OrderRefundForm']
+
+
+def validate_amount(form, field):
```
jace: If it's only used once, it can be a method on the form.

boxoffice/views/order.py
```diff
@@ -541,22 +543,17 @@ def cancel_line_item(line_item):
...
-        if (order.refunded_amount + requested_refund_amount) > order.paid_amount:
-            return api_error(message='Invalid refund amount, must be lesser than net amount paid for the order',
-                    status_code=403,
-                    errors=['excess partial refund'])
+    if request_method == 'GET':
+        return jsonify(form_template=render_form(form=form, title=u"Partial refund", submit=u"Refund", with_chrome=False))
 
-        payment = OnlinePayment.query.filter_by(order=order,
-            pg_payment_status=RAZORPAY_PAYMENT_STATUS.CAPTURED).one()
+    if form.validate():
```
jace: As general safety principle, I'd say always use `validate_on_submit` if you're expecting a non-GET request. The previous block trapping `GET` may be moved in future code refactoring.

PR #218, opened 2018-04-20
Allow navigating to an order page with a receipt no.

shreyas-satish:
- Adds ability to navigate to a specific order, by entering a receipt no. in the text box on the organization page.
- Provides individual URLs for orders


boxoffice/views/admin_order.py
```diff
@@ -80,7 +80,48 @@ def admin_orders(item_collection):
...
+    }
+    return jsonify(
+        org=org,
+        ic=ic,
+        order=order,
+        line_items=format_line_items(order_dict['order'].line_items)
+    )
+
+
+@app.route('/admin/o/<org_name>/order/<receipt_no>', methods=['GET', 'POST'])
```
jace: Use `<int:receipt_no>`

boxoffice/views/admin_order.py
```diff
@@ -80,7 +80,48 @@ def admin_orders(item_collection):
...
+@lastuser.requires_permission(('siteadmin', 'org_admin'))
+@lastuser.requires_login
+@render_with({'text/html': 'index.html.jinja2', 'application/json': jsonify_order})
+def admin_org_order(org_name, receipt_no):
+    org = Organization.query.filter_by(name=org_name).one_or_404()
+    try:
+        int(receipt_no)
+    except:
+        abort(400)
```
jace: This can be removed if you move it to the URL, but you may have to check what happens if the parameter is not an integer. My guess, you'll get a 404 instead of a 400.

boxoffice/views/admin_order.py
```diff
@@ -80,7 +80,48 @@ def admin_orders(item_collection):
...
+    order = Order.query.filter(Order.organization == org, Order.receipt_no == receipt_no).one_or_404()
```
jace: The `org` and `order` queries can be moved to `load_models` if int handling goes to the URL. You could also reduce these two queries into a single query:

`order = Order.query.join(Organization).filter(Order.receipt_no == receipt_no, Organization.name == org_name).one_or_404()`

boxoffice/views/admin_order.py
```diff
@@ -80,7 +80,48 @@ def admin_orders(item_collection):
...
+    line_items = LineItem.query.filter(LineItem.order == order,
+        LineItem.status.in_([LINE_ITEM_STATUS.CONFIRMED, LINE_ITEM_STATUS.CANCELLED])).all()
```
jace: Actually, all three queries can be combined into one:

```python
LineItem.query.join(Order).join(Organization).filter(
    LineItem.status.in_([LINE_ITEM_STATUS.CONFIRMED, LINE_ITEM_STATUS.CANCELLED]),
    Order.receipt_no == receipt_no,
    Organization.name == org_name
    ).one_or_404()
```

shreyas-satish:
@jace Thanks, using the `int` prefix in the url route and also switched to `load_models`. I'll need to load the line items though, to display them.

PR #219, opened 2018-04-23
Added ability to delete discount policy

shreyas-satish:


boxoffice/views/admin_discount.py
```diff
@@ -152,13 +154,40 @@ def admin_edit_discount_policy(discount_policy):
...
+            errors=form.errors)
+
+    db.session.delete(discount_policy)
+    db.session.commit()
+    return api_success(result={}, doc="Discount policy deleted.", status_code=200)
+
+
+@app.route('/admin/o/<org_name>/discount_policy/<discount_policy_id>/delete', methods=['GET', 'POST'])
+@lastuser.requires_login
+@render_with({'text/html': 'index.html.jinja2', 'application/json': jsonify_delete_discount_policy})
```
jace: Is this really meant to render with `index.html.jinja2`?

boxoffice/views/admin_discount.py
```diff
@@ -152,13 +154,40 @@ def admin_edit_discount_policy(discount_policy):
             errors=discount_policy_form.errors)
 
 
+def jsonify_delete_discount_policy(discount_policy_dict):
```
jace: This function is view logic. It is not a renderer. It should be a part of the view function.

boxoffice/views/admin_discount.py
shreyas-satish: Yes, that's the boot up page for the SPA.

boxoffice/views/admin_discount.py
shreyas-satish: Will renaming the function to `delete_discount_policy` work? It's been extracted out to make it work with `render_with`.

boxoffice/views/admin_discount.py
jace: A comment to that effect will help.

boxoffice/views/admin_discount.py
jace: `render_with` is the wrong tool here. Flask will let you specify separate handlers by content type, so use that and make separate views.

PR #223, opened 2018-04-30
Apply to attend

shreyas-satish:
Adds the ability for an event manager to control access to specific tickets with the help of discount coupons.

To use:
- For an item that needs access control, check the `restricted_entry` flag in the item form.
- Create a discount policy for the said item and generate coupons to grant access.


boxoffice/models/discount_policy.py
```diff
@@ -165,6 +165,20 @@ def line_items_count(self):
         from ..models import LineItem, LINE_ITEM_STATUS
         return self.line_items.filter(LineItem.status == LINE_ITEM_STATUS.CONFIRMED).count()
 
+    @classmethod
+    def is_valid_access_coupon(cls, item, code_list):
```
jace: This method needs documentation on why there's a search through both `DiscountPolicy` and `DiscountCoupon` models.

boxoffice/views/item_collection.py
```diff
@@ -1,35 +1,43 @@
...
-from boxoffice.models import Organization, ItemCollection, Item
-from utils import xhr_only, cors
+from boxoffice.models import Organization, ItemCollection, Item, DiscountPolicy
+from utils import xhr_only, cors, sanitize_coupons
 from boxoffice.data import indian_states
 
 
 def jsonify_item(item):
+    if item.restricted_entry:
+        code_list = request.args.get('code') and sanitize_coupons(json.loads(request.args.get('code')))
```
jace: Why are we accepting JSON in a GET parameter? If you need to accept a list, use `request.args.getlist('code')`

shreyas-satish: Changed to use `request.args.getlist('code')`.

boxoffice/models/discount_policy.py
shreyas-satish: Added documentation.

boxoffice/views/item_collection.py
```diff
@@ -1,35 +1,42 @@
...
-from boxoffice.models import Organization, ItemCollection, Item
-from utils import xhr_only, cors
+from boxoffice.models import Organization, ItemCollection, Item, DiscountPolicy
+from utils import xhr_only, cors, sanitize_coupons
 from boxoffice.data import indian_states
 
 
 def jsonify_item(item):
+    if item.restricted_entry:
+        code_list = request.args.getlist('code') and sanitize_coupons(request.args.getlist('code'))
```
jace: This line is confusing. It could be: `code_list = sanitize_coupons(request.args.getlist('code')) if 'code' in request.args else None`

vidya-ram: Changed

PR #294, opened 2019-07-16
Unique coupon code

### LINKED ISSUE #290: Uniqueness of coupon code
Right now `DiscountCoupon.code` are unique only inside the same `DiscountPolicy`.  But here [in this bit](https://github.com/hasgeek/boxoffice/blob/9f16e65e633bb57107552ba5c65113a72df90c2a/boxoffice/models/discount_policy.py#L168) we assume that coupon codes are unique globally. Which caused couple of failures when Mak created couple of coupon codes with same name as some old coupon codes.

So there is a confusion about the uniqueness of coupon codes. Are they unique -

1. within a discount policy
2. across multiple discount policies but within a single item (as `Item` links to multiple discount policies)
3. globally, across all discount policies

We'll need to update our logic accordingly.

/cc @jace @vidya-ram

iambibhas:
Fixes #290.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,24 @@
+[run]
+source = boxoffice
+
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # Have to re-enable the standard pragma
+    pragma: no cover
+
+    # Don't complain about missing debug-only code
+    def __repr__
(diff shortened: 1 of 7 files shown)


iambibhas:
This is not the right solution. This will need to be a much bigger PR.

PR #296, opened 2019-07-18
Move webpack config and package json to assets folder. Add Make file.

vidya-ram:


.gitignore
```diff
@@ -16,3 +16,5 @@ boxoffice/static/gen/*
 *.sql
 *.sql.gz
 *.sql.bz2
+node_modules
```
jace: The ignore line above for `static/js/node_modules` can be removed.

Issue #104, opened 2016-09-15
Incase of transfer of ticket, email confirmation requires to be sent to previous assignee


vidya-ram:

vidya-ram:
Fixed in #297

PR #297, opened 2019-08-01
Ticket transfer deadline and reassignement notification

### LINKED ISSUE #244: Send notification when ticket is re-assigned
When a ticket is assigned to an individual, they receive an email confirming the assignment (do they?). However, when the ticket is re-assigned to someone else, the previous assignee does not get a notice about no longer having a ticket. This notice must be sent.

(Reported by @makxy based on observed user behaviour.)

### LINKED ISSUE #301: GST Invoice Options
### Background
There are 2 types of GST Invoices: 

1. B2B invoices where the buyer and seller are both businesses (Business to Business) and
2. B2C invoices where the buyer is a retail consumer. 

The distinguishing factor between these two invoices is that only businesses have a GST number. Currently Box office provides the following information as part of the invoices section:
```json
"invoices" : [
      {
         "city" : "Hyderabad",
         "status" : "Draft",
         "invoiced_at" : "Thu, 20 Jun 2019 05:44:29 GMT",
         "street_address_2" : null,
         "state_code" : "TG",
         "invoice_no" : 20000,
         "country_code" : "IN",
         "street_address_1" : "Address",
         "invoicee_company" : "XYZ",
         "buyer_taxid" : null,
         "state" : null,
         "invoicee_email" : "name@XYZ.com",
         "seller_taxid" : null,
         "postcode" : "500032"
      }
```

The key field to note is the buyer_taxid. If it is null, then it is a B2C invoice or it contains a valid GST#, then it is a B2B invoice.  B2B invoices are thus always issued in the name of the company (invoicee_company) and B2C invoices are always issued in the name of the individual who bought it. 

### Reimbursements
Field experience indicates that most of the participants don't fill the GST#  in the invoice section, but do fill the invoicee_company name properly. This presents a problem in issuing invoices because to claim reimbursements from their employers, participants need the company name in the invoices and also their names. So there is a need to distinguish the reimbursement case correctly in the UI and also report it in the invoices JSON (reimbursement: true or equivalent) as shown below: 

```json
"invoices" : [
      {
         "city" : "Hyderabad",
         "status" : "Draft",
         "invoiced_at" : "Thu, 20 Jun 2019 05:44:29 GMT",
         "street_address_2" : null,
         "state_code" : "TG",
         "invoice_no" : 20000,
         "country_code" : "IN",
         "street_address_1" : "Address",
         "invoicee_company" : "XYZ",
         "buyer_taxid" : null,
         "state" : null,
         "invoicee_email" : "name@XYZ.com",
         "seller_taxid" : null,
         "postcode" : "500032", 
         "reimbursement": true
      }
```

While the exact semantics might not be the same as reported above, what is clear is the need to handle reimbursement as a separate item in the B2C invoice case.

iambibhas:
This includes two fixes -

1. Adds a `transferable_until` field to `Item` to mention till what time can tickets in that item be transferred to someone else.
2. Sends notification to the old attendee when a ticket is transferred.

Fixes #301 
Fixes #244, #104


boxoffice/templates/ticket_reassignment_mail.html.jinja2
```diff
@@ -87,8 +87,8 @@
           </td>
 
           <td class="details-wrapper">
-            <p>Your ticket has been transfered to {{ new_assignee.fullname }}, by {{ current_auth.user.fullname }}.</p>
-            <p>If you believe this was because of an error, please reach out to us at <a href="mail:support@hasgeek.com">support@hasgeek.com</a>.</p>
+            <p>Your ticket has been transfered to {{ new_assignee.fullname }}{% if current_auth.user %}, by {{ current_auth.user.fullname }}{% endif %}.</p>
```
jace: This needs `{% trans %}` tags (in future if not now). That means the text has to be a single sentence and the `if` statement has to be outside the trans tag.

boxoffice/views/participant.py
```diff
@@ -64,11 +67,22 @@ def assign(order):
             line_item.current_assignee.details = assignee_details
             db.session.commit()
         else:
+            if not line_item.is_transferable and line_item.current_assignee is not None:
```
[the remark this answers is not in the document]
jace: Actually, we want this:
```python
if line_item.current_assignee is None or line_item.is_transferable:
    # Allow
else:
    # Deny
```
Since there is no Allow block, we can get to the Deny block by applying a `not`:
```python
if not (line_item.current_assignee is None or line_item.is_transferable):
```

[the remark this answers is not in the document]
jace: Previous incident of clarity-based if/else constructs: https://github.com/hasgeek/coaster/blob/5952d8eb7ed51b8405ace6d82ab71843abe971e2/coaster/sqlalchemy/immutable_annotation.py#L59-L66

There's one other example somewhere I'm unable to find. We had an elaborate if condition broken down into multiple if/pass statements with a final else.

boxoffice/views/participant.py
```diff
@@ -64,11 +67,22 @@ def assign(order):
...
+                return (
+                    {
+                        'status': 'error',
+                        'error': 'ticket_not_transferable',
+                        'error_description': u"Ticket transfer deadline is over. It can no longer be transfered",
```
jace: Requires `_(...)`

boxoffice/views/participant.py
```diff
@@ -77,9 +91,16 @@ def assign(order):
...
             db.session.add(new_assignee)
             db.session.commit()
             send_ticket_assignment_mail.queue(line_item.id)
+
+            if old_assignee is not None:
+                # Send notification to previous assignee
+                send_ticket_reassignment_mail.queue(
+                    line_item.id, old_assignee.id, new_assignee.id
+                )
         return {'status': 'ok', 'result': {'message': 'Ticket assigned'}}
```
jace: Also requires `_(...)`

jace: Unless the front-end is treating 'Ticket assigned' as a literal string. @vidya-ram?

vidya-ram: This string is not used in the front-end

boxoffice/templates/ticket_reassignment_mail.html.jinja2
```diff
@@ -0,0 +1,126 @@
...
+          </p>
+
+          <p>
+            {% if old_assignee.email != order.buyer_email %}
+              {% trans buyer_email=order.buyer_email, buyer_fullname=order.buyer_fullname %}
+                If you believe this was because of an error, please reach out to <a href="mail:{{ buyer_email }}">{{ buyer_fullname }}</a>.
+              {% endtrans %}
+            {% else %}
+              {% trans %}
+                If you believe this was because of an error, please reach out to <a href="mail:support@hasgeek.com">support@hasgeek.com</a>.
```
vidya-ram: This needs to be `{{order.item_collection.organization.support_email}}`

boxoffice/models/line_item.py
```diff
@@ -152,6 +156,18 @@ def confirm(self):
...
+    @property
+    def is_assignable(self):
+        return utcnow().date() <= self.item.event_date
+
+    @property
+    def is_transferable(self):
+        if self.current_assignee is None:
+            return self.is_assignable
+        else:
+            return utcnow() <= self.item.transferable_until if self.item.transferable_until is not None else self.is_assignable
```
jace: Should be `< self.item.transferable_until`. If you specify the transfer cut-off as midnight, then it should work until midnight but not at midnight.

boxoffice/templates/invoice_form.html.jinja2
```diff
@@ -18,7 +18,7 @@
   </div>
 
   <div class="row" id="boxoffice-invoice">
-    <p class="text-center" id="notify-msg"><span id="error-description">Loading. </span>If this takes too long, please mail {{order.item_collection.organization.support_email}} with your receipt number.</p>
+    <p class="text-center" id="notify-msg"><span id="error-description">Loading. </span>If this takes too long, please mail {{ org.details.get('support_email', '') }} with your receipt number.</p>
```
jace: Template needs trans tags.

boxoffice/templates/order.html.jinja2
```diff
@@ -21,7 +21,7 @@
     </div>
     
     <div class="row" id="boxoffice-order">
-      <p class="text-center" id="notify-msg"><span id="error-description">Loading. </span>If this takes too long, please mail {{order.item_collection.organization.support_email}} with your receipt number.</p>
+      <p class="text-center" id="notify-msg"><span id="error-description">Loading. </span>If this takes too long, please mail {{ org.details.get('support_email', '') }} with your receipt number.</p>
```
jace: Template needs trans tags.

boxoffice/templates/ticket_reassignment_mail.html.jinja2
```diff
@@ -0,0 +1,126 @@
...
+          </p>
+
+          <p>
+            {% if old_assignee.email != order.buyer_email %}
+              {% trans buyer_email=order.buyer_email, buyer_fullname=order.buyer_fullname %}
+                If you believe this was because of an error, please reach out to <a href="mailto:{{ buyer_email }}">{{ buyer_fullname }}</a>.
+              {% endtrans %}
+            {% else %}
+              {% trans support_email = org.details.get('support_email', '') %}
+                If you believe this was because of an error, please reach out to <a href="mailto:{{ support_email }}">{{ support_email }}</a>.
```
jace: If we are sharing the buyer's email with the assignee, then we should warn in the buyer in the assignment form. Or remove the email here.

boxoffice/templates/ticket_reassignment_mail.html.jinja2
```diff
@@ -0,0 +1,126 @@
...
+            {% endif %}
+          </p>
+          </td>
+        </tr>
+      </table>
+    </div>
+
+    <div class="terms-content">{{ org.details.get('ticket_faq', '') | safe }}</div>
+
+    <p>{% trans %}Thank you{% endtrans %},</p>
```
jace: Comma should not be outside trans.

boxoffice/templates/ticket_reassignment_mail.html.jinja2
iambibhas: How do we warn the buyer? Because i think it happens often that someone buys all the tickets and then assigns them one by one. and the asignees get mail. but then if their ticket gets reassigned to someone else without their own consent, what course of action do they have?

boxoffice/models/line_item.py
```diff
@@ -152,6 +156,18 @@ def confirm(self):
     def current_assignee(self):
         return self.assignees.filter(Assignee.current == True).one_or_none()  # NOQA
 
+    @property
+    def is_assignable(self):
+        return utcnow().date() <= self.item.event_date
```
jace: I just realised this will blow on day 2 of the event, for someone who only shows up on the second day. Are events on a per-day basis?

Also, `utcnow().date()` should be cast into event timezone first. Is that recorded?

boxoffice/templates/ticket_reassignment_mail.html.jinja2
jace: Whoever has access to the assignment form should have a notice displayed: when you assign this ticket, the buyer's email will be shared. Or we should remove the email here. The name alone should be sufficient to help the assignee recognise the assigner.

iambibhas: Alright. I'll remove the email and keep the name.

boxoffice/models/line_item.py
iambibhas: There is no timezone stored anywhere in boxoffice. There are only the timezone in app config and the user's timezone.

PR #307, opened 2019-11-18
Set 0 as value for empty label

iambibhas:


boxoffice/forms/item.py
```diff
@@ -82,18 +82,24 @@ class ItemForm(forms.Form):
...
+    def validate_place_supply_state_code(self, field):
+        if field.data <= 0:
+            # state short codes start from 1,
+            # and 0 means empty value as mentioned above in set_queries
+            raise forms.ValidationError(__("Please select a state"))
+
     def validate_transferable_until(self, field):
         if field.data and field.data.date() > self.event_date.data:
-            raise forms.ValidationError("Ticket transfer deadline cannot be after event date")
+            raise forms.ValidationError(__("Ticket transfer deadline cannot be after event date"))
```
jace: Single `_` here as this is at runtime. Not a breaker though.

PR #319, opened 2020-03-12
assignee email is unique per item

iambibhas:


boxoffice/forms/participant.py
```diff
@@ -22,7 +22,8 @@ class AssigneeForm(forms.Form):
 
     def validate_email(self, field):
         existing_assignees = (
-            Assignee.query.filter(LineItem.order == self.edit_parent.order)
+            Assignee.query.join(LineItem, Item, Order)
```
jace, line 25: This should be three joins. The second parameter to join is the join clause.

jace (changes requested):

PR #352, opened 2021-02-25
Raise an JS event with ticket prices and quantity

vidya-ram:


boxoffice/views/item_collection.py
```diff
@@ -85,7 +85,7 @@ def item_collection(item_collection):
         category_json = jsonify_category(category)
         if category_json:
             categories_json.append(category_json)
-    return jsonify(html=render_template('boxoffice.html.jinja2'), categories=categories_json, refund_policy=item_collection.organization.details.get('refund_policy', ''))
+    return jsonify(html=render_template('boxoffice.html.jinja2'), categories=categories_json, refund_policy=item_collection.organization.details.get('refund_policy', ''), currency=CURRENCY_SYMBOL['INR'])
```
jace, line 88: Isn't currency part of the model?

vidya-ram: It seems to be part of payment model

PR #361, opened 2021-05-09
Apply pre-commit config from Funnel and refactor code to comply

jace:


vidya-ram (commented):
boxoffice/templates/ractive.js is included in boxoffice.js.jinja2

PR #445, opened 2023-05-11
Modernize code to pass linter checks

jace:
Also drop the obsolete html5print dependency.


jace:
Finally passing all pre-commit checks.

jace:
This still has some way to go before Boxoffice can be merged into Funnel:

1. The dependencies need to be a subset of what Funnel uses
2. The User and Organization models should not have any additional fields, as Funnel's Account won't be getting them
3. The `scripts/` folder needs to be dropped, so whatever useful functionality it still has needs to be moved into a CLI command

PR #473, opened 2023-11-22
Send order purchase messages to telegram

djamg:


boxoffice/messageclient.py
```diff
@@ -0,0 +1,12 @@
+import requests
+
+from . import app, rq
+
+
+@rq.job('boxoffice')
+def send_telegram_message(order):
+    with app.test_request_context():
+        message_text = f'{order.buyer_fullname} purchased {order.line_item.item.title}'
```
jace, line 9: Gettext needed here.

boxoffice/messageclient.py
```diff
@@ -0,0 +1,12 @@
...
+        send_text = f'https://api.telegram.org/bot{app.config["TELEGRAM_BOT_TOKEN"]}/sendMessage'
+        params = {'chat_id': app.config["TELEGRAM_CHAT_ID"], 'text': message_text}
```
jace, line 11: Optional `thread_id` needed here.

jace, line 11: Single quotes for config dict.

boxoffice/views/order.py
```diff
@@ -438,6 +439,7 @@ def payment(order):
                 invoice_no=order.invoice_no,
             ),
         )
+        send_telegram_message.queue(order=order)
```
jace, line 442: Never pass an object to an RQ job. It uses pickles to pass data and that's a security risk. Pass a simple data type like an int. See the previous line for an example.

jace (changes requested):
A couple minor changes needed.

boxoffice/messageclient.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@rq.job('boxoffice')
+def send_telegram_message(buyer_fullname, line_item_title):
+    with app.test_request_context():
+        message_text = _(f'{buyer_fullname} purchased {line_item_title}')
+        send_text = f'https://api.telegram.org/bot{app.config["TELEGRAM_BOT_TOKEN"]}/sendMessage'
+        params = {
+            'chat_id': app.config['TELEGRAM_CHAT_ID'],
+            'message_thread_id': app.config['TELEGRAM_MESSAGE_THREAD_ID'],
```
jace, line 15: This is optional, so use `app.config.get('TELEGRAM_THREADID')`

Note: the telegram key names we're using in Funnel are `TELEGRAM_APIKEY`, `TELEGRAM_CHATID` and `TELEGRAM_THREADID`.

boxoffice/messageclient.py
```diff
@@ -0,0 +1,18 @@
...
+
+from baseframe import _
+
+from . import app, rq
+
+
+@rq.job('boxoffice')
+def send_telegram_message(buyer_fullname, line_item_title):
+    with app.test_request_context():
+        message_text = _(f'{buyer_fullname} purchased {line_item_title}')
```
jace, line 11: `f` strings can't be used inside `_()` as it will send the rendered string for translation. Change to:

```python
message_text = _("{user} purchased {title}").format(user=buyer_fullname, title=line_item_title)
```

boxoffice/views/order.py
```diff
@@ -438,6 +439,11 @@ def payment(order):
                 invoice_no=order.invoice_no,
             ),
         )
+        if app.config['TELEGRAM_BOT_TOKEN']:
```
jace, line 442: Check for both API key and chat id before queueing this job.

jace (changes requested):

boxoffice/views/order.py
```diff
@@ -438,6 +439,11 @@ def payment(order):
                 invoice_no=order.invoice_no,
             ),
         )
+        if app.config['TELEGRAM_APIKEY'] and app.config['TELEGRAM_CHATID']:
+            send_telegram_message.queue(
+                buyer_fullname=order.buyer_fullname,
+                line_item_title=order.line_item.item.title,
```
jace, line 445: `order.line_items` is a list. Don't have this extraction logic here. Pass just `order.id` to the background job, and let is extract the data it needs.

jace (changes requested):

boxoffice/messageclient.py
```diff
@@ -0,0 +1,24 @@
+import requests
+
+from baseframe import _
+
+from . import app, rq
+from .models import Order
+
+
+@rq.job('boxoffice')
+def send_telegram_message(order_id):
```
jace, line 10: Add types to signature: `def send_telegram_message(order_id: int) -> None:`

boxoffice/messageclient.py
```diff
@@ -0,0 +1,24 @@
...
+    with app.test_request_context():
+        order = Order.query.get(order_id)
```
jace, line 12: `order` may be `None`, so add `if order is not None:` block for the rest of the code.

boxoffice/messageclient.py
```diff
@@ -0,0 +1,24 @@
...
+        message_text = _("{user} purchased {title}").format(
+            user=order.buyer_fullname, title=order.line_item.item.title
+        )
```
jace, line 15: Model `Order` has no attribute named `line_item`. It has a list named `line_items`. You'll have to get the title from each item in the list and construct a larger message with the full order.

jace (changes requested):
The `line_items` change is pending.

boxoffice/views/order.py
```diff
@@ -438,6 +439,8 @@ def payment(order):
                 invoice_no=order.invoice_no,
             ),
         )
+        if app.config['TELEGRAM_APIKEY'] and app.config['TELEGRAM_CHATID']:
```
jace, line 442: This will fail if those keys are missing in config. Change to `if app.config.get('TELEGRAM_APIKEY') and app.config.get('TELEGRAM_CHATID')`

PR #515, opened 2024-06-27
Fix discount policy form and ticket assignment page

vidya-ram:
1. Fix Discount Type input values in the new/edit discount policy form
2. Fix display of ticket description in the ticket assignment form


jace (approved):
This should be an enum object that's kept in sync with backend, or it should be string-based. Can you check for other enums hardcoded like this? We're now using 1-indexed enums in the backend.

